PR #11, opened 2021-03-03
Stable 0.2.4

nategbrown9:


androidApp/build.gradle.kts
```diff
@@ -17,8 +17,8 @@ repositories {
 dependencies {
     implementation("org.jetbrains.kotlin:kotlin-stdlib:${Versions.kotlin}")
     implementation(Deps.AndroidX.annotation)
-    implementation(project(":bridgeClient"))
-    implementation(project(":bridgeClientPresentation"))
+    implementation(project(":bridge-client"))
+    implementation(project(":bridge-client-presentation"))
```
liujoshua, line 21: Do you think we'll ever have an iOS presentation library in this repo?

liujoshua (approved):

androidApp/build.gradle.kts
nategbrown9: Maybe at some point, but initially I would guess it would be some sort of swift package. I debated calling this bridge-client-presentation-android, but maybe longterm there will be a multiplatform UI framework that will support both in one module.

PR #13, opened 2021-03-18
Add call to complete upload session

nategbrown9:
This also has the changes to enable publishing of the presentation module.


liujoshua (approved):
Code looks good to merge. 

I was mulling over whether we should include this in our new client library. A couple years ago, I did some work on the server end to trigger upload completion based on S3 notifications. This call _should_ now be redundant and result in a no-op on Bridge

Erin-Mounts:
@liujoshua fyi it's my understanding (from @DwayneJengSage  I believe?) that it turns out that S3 notifications aren't 100% reliable and if we need it to be more bulletproof then a belt-and-suspenders approach to try to catch it from both ends might be worth the trouble. AFAIK, FWIW, YMMV, etc.

PR #14, opened 2021-04-07
Test cleanup

nategbrown9:
- Adds a github action to run tests on PRs
- Moves some tests to commonTest so they run on both Android and iOS
- Adds an AuthenticationIntegrationTest that runs against Bridge server


syoung-smallwisdom (approved):
looks good?

PR #15, opened 2021-04-09
Add assessmentmodel-sdk module

liujoshua:


syoung-smallwisdom (changes requested):
I'm pretty sure that this is adding the AssessmentModel to the common crossplatform code. Unfortunately, we can't do that or it will break iOS module structure. This needs to be added as a wrapper in the android code instead.

nategbrown9:
> I'm pretty sure that this is adding the AssessmentModel to the common crossplatform code. Unfortunately, we can't do that or it will break iOS module structure. This needs to be added as a wrapper in the android code instead.

This is adding a new assessmentmodel-sdk module that depends on bridge-client and AssessmentModel.

syoung-smallwisdom (approved):

nategbrown9 (approved):
Looks good. Could remove the wizard generated Platform.kt and Greeting.kt files.

PR #16, opened 2021-04-09
Feature/upload

liujoshua:
Builds on top of https://github.com/Sage-Bionetworks/BridgeClientKMM/pull/15

Not able to get new version of upload/archive working yet, so depending on old code


syoung-smallwisdom (commented):
So... Does this add the AssessmentModel to the *crossplatform* library? Sorry, I don't really understand gradles.

Unfortunately, we can't do that and use this on iOS because that would require building the AssessmentModel in the same module as BridgeClient (messes with imports).

nategbrown9 (approved):

PR #17, opened 2021-04-14
Set up Swift PM

syoung-smallwisdom:
When adding code to the iosApp.xcodeproj you will want to add the wrapper files into the BridgeClientUI group which points at the shared folder used to build that code as a Swift module. Currently, the project and Swift package are only set up to allow using SageResearch b/c the AssessmentModel-KotlinNative repo is not currently set up to build the xcframework automagically. This set up assumes that before any of this is used in production code, those pipelines will be set up so that the iOS framework has parity with Android.

Note: Additionally, I did *not* attempt to set up for using Carthage dependency management so this requires MTB to use SwiftPM.


Erin-Mounts (approved):
👍

nategbrown9 (approved):
Looks good.

PR #18, opened 2021-04-23
Participant schedule timeline

nategbrown9:
This includes an on demand, in memory, extraction of today's scheduled sessions from the timeline. This should be plenty fast for the schedules needed by MTB in the near term, and unblocks us to start building UI. I am holding off on building a better caching mechanism until we have the history api.


syoung-smallwisdom (approved):
looks good

PR #27, opened 2021-06-08
Set up koin for iOS

syoung-smallwisdom:
This is *very* lightly tested against iOS


SwiftPackage/Sources/BridgeClientUI/BridgeClientAppViewModel.swift
```diff
@@ -0,0 +1,101 @@
...
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import SwiftUI
+import BridgeClient
+
+open class BridgeClientAppViewModel : ObservableObject {
```
nategbrown9: Not sure what to call this, but I don't think it's a ViewModel. It is tied to the app as a whole not an individual view or screen.

nategbrown9 (approved):
Looks good. Just wondering if we can come up with a better name for BridgeClientAppViewModel.

PR #28, opened 2021-06-11
Update getSessionsForToday

nategbrown9:
- No longer need to pass in activity events list 
- Will emit new value anytime cached values for events, timeline, or adherence records change
- Should only emit a value when there really was a change.


syoung-smallwisdom (approved):
looks good

PR #29, opened 2021-06-11
Timeline ios

syoung-smallwisdom:
Added the timeline manager. At this point it is *only* getting "today's" timeline and there's no logic for refreshing on "new day" or really a sense for what "new day" means. Therefore, on iOS we're going to need to figure something out b/c objects can live for a very long time. 

mañana ¯\_(ツ)_/¯ 

Anyhow, this lets me show a schedule. :)

### DIFF
--- a/bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
+++ b/bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
@@ -21,20 +21,15 @@ class NativeTimelineManager(
     private val viewUpdate: (List<NativeScheduledSessionWindow>) -> Unit
 ) : KoinComponent {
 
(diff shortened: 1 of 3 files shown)


bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -0,0 +1,93 @@
...
+        scope.cancel()
+        observeTodaySchedule()
+    }
+
+    fun onCleared() {
+        scope.cancel()
+    }
+}
+
+fun LocalDateTime.toNative() : NSDate? {
```
nategbrown9: if you add `import kotlinx.datetime.toNSDateComponents` you could simplify this to: `return this.toNSDateComponents().date`

nategbrown9 (approved):
Looks good.

PR #31, opened 2021-06-14
Upload expired sessions

nategbrown9:
This allows partially completed assessment results to be saved for upload, but only uploaded when the scheduled session has expired. If a user finishes the assessment before the expiration time, the partially one will be deleted and the finished one uploaded.


syoung-smallwisdom (approved):
looks good

PR #33, opened 2021-06-18
Archive uploads

syoung-smallwisdom:
WIP for uploading archives.

This is includes porting all the BridgeApp code for uploading V2 archives (but not the BridgeSDK *actual* uploading stuff).

It also moves some of the timeline view model stuff into this code. I decided to do that b/c it was starting to get complicated and I learned yesterday that Kotlin objects are not thread-safe so I have to have the model create Swift objects to pass around in order to do the archive processing on a background thread.

I also learned that SwiftUI just cannot handle rendering cascading if statement and so end up using a lot of state variables that have to be set on a thing called the `TimelineSession`. *sigh* Good times.


nategbrown9 (approved):
Looks good.

PR #34, opened 2021-06-23
UI/UX tweeks

syoung-smallwisdom:
So... still learning about SwiftUI lifecycle and rendering loops. This is to support changes in the app code.


nategbrown9 (approved):
Looks good.

PR #35, opened 2021-06-24
Schedules with next up

syoung-smallwisdom:
I figure may as well *not* optimize until we need to so I am just getting all the future schedules and then filtering out the schedules for the "next up" day. 

This also can be used on iOS to set up notifications at a later time.

Additionally, I "Kotlined" the code some. Mostly just fixed warnings about get(0) and cleaned up the extraction function to make it more readable. FYI, it's better to use `forEach` and then exit early with `return@forEach` or to use`map` and `flatten` when running this kind of stuff. Makes it easier to extract, refactor, and read.


nategbrown9 (approved):
Looks good.

PR #36, opened 2021-06-29
Added changes to shared code for supporting notifications on iOS

syoung-smallwisdom:
I *think* this provides me with all I will need to be able to support notifications on iOS. I also did some additional cleanup of the filtering code to always return "next day" sessions even if the design calls for them being hidden in the "today" view.  Ideally, future changes should be additive only.

### DIFF
--- a/SwiftPackage/Sources/BridgeClientUI/ObservableObjects/TodayTimelineViewModel.swift
+++ b/SwiftPackage/Sources/BridgeClientUI/ObservableObjects/TodayTimelineViewModel.swift
@@ -40,7 +40,7 @@ open class TodayTimelineViewModel : NSObject, ObservableObject {
     
     @Published open var today: Date = Date()
     @Published open var studyId: String?
-    @Published open var sessions: [TimelineSession] = []
+    @Published open var sessions: [TimelineSession] = [] 
     
     /// A flag that can be used to set whether or not a view is presenting the assessment. How the assessment is
     /// presented is up to the views built specifically for a given application.
@@ -59,7 +59,7 @@ open class TodayTimelineViewModel : NSObject, ObservableObject {
         
     private var schedules: [NativeScheduledSessionWindow] = [] {
         didSet {
-            let newSessions: [TimelineSession] = schedules.map { schedule in
+            let newSessions: [TimelineSession] = schedules.compactMap { schedule in
                 guard let existingSession = self.sessions.first(where: { $0.instanceGuid == schedule.instanceGuid })
                 else {
                     return TimelineSession(schedule)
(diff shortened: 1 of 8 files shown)


SwiftPackage/Sources/BridgeClientUI/ObservableObjects/TodayTimelineViewModel.swift
```diff
@@ -40,7 +40,7 @@ open class TodayTimelineViewModel : NSObject, ObservableObject {
     
     @Published open var today: Date = Date()
     @Published open var studyId: String?
-    @Published open var sessions: [TimelineSession] = []
+    @Published open var sessions: [TimelineSession] = []
```
Erin-Mounts:
```suggestion
    @Published open var sessions: [TimelineSession] = []
```

Erin-Mounts (approved):
One nit, otherwise looks good to me

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -258,18 +321,47 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+
+        // Exit early if expired and not completed.
+        if (!filterType.includeExpired() && isExpired && !isCompleted) {
+            return null
+        }
+
+        // If we are filtering for today schedules and one was found that is available today
+        // but *not* available NOW then track that and do not include future dates.
+        if (rulesForAvailableToday() && startInstant > instantInDay) {
+            foundState.found = true
```
nategbrown9: Won't this prevent alwaysIncludeNextDay from working when we have a session today that isn't available yet? Also if we have sessions keying off of different events, we could end up with an incomplete subset of nextDay sessions.

syoung-smallwisdom: Good catch. I hadn't realized that my test case `testScheduledSessionsDay1TodayAndTomorrow()` is for `TodayAndNotifications` and all the test schedules have notifications on them.  

WRT incomplete subset for next day, if a session keys off of an event that has not been set yet (for example, finishing the session that is currently available or is available later today) then correct, it will not be included in the "next day" events *but* it also doesn't matter because if and when the adherence record is updated that triggers an event update, there is a new set of schedules, this thing will update the UI accordingly. Including the next day schedules is really intended on iOS for the scenario where the participant does *not* do anything that would trigger a change. They close the app and come back later that day when the schedules have *not* changed but the UI/UX rules have. This allows the UI to filter the schedules without waiting for the database/server stuff to catch up. In other words, the view is presented in the state it *should* be shown in rather than the state as determined by the server.

syoung-smallwisdom: Fixed

nategbrown9 (approved):
Looks good.

PR #37, opened 2021-07-01
Fix get sessions

nategbrown9:
Even though the result from it is currently unused, we need to include a call to get adherence records in the combine portion of getSessionsForToday. This will trigger the resulting flow to emit a new value anytime the adherence records cache is updated.


liujoshua (approved):

nategbrown9:
Shannon already fixed this in another PR, but the comment in the code is still useful.

PR #38, opened 2021-07-06
Update sage research 4.1.5

syoung-smallwisdom:
Note: None of this is used by the MTB app right now with the exception of changing `RSDResult` to `ResultData` b/c that app is not set up to be able to use SwiftPM, but I wanted to keep the package up-to-date as documentation.


Erin-Mounts (approved):
looks good to me

PR #39, opened 2021-07-06
Build xcframeworks

syoung-smallwisdom:
I don't totally understand why this works or what the original code was trying to do, but it wasn't working when attempting to build on device or archive the app for distribution.

I need to be able to build an XCFramework so that I can build a release configuration and so that I can build debug for both device and simulator. These changes set up a target that builds for both the Swift package and against the MTB app where I embed the xcframework directly. The built XCFramework was able to access the database to get the auth token, schedules, and adherence records that were already cached both on device and using the simulator. I was able to archive the XCFramework. 

So... I guess it works?


nategbrown9 (approved):
Looks good?

PR #40, opened 2021-07-08
Add back in work-around to allow building for Android

syoung-smallwisdom:
I’m not sure why the iOS framework is built as a part of building for Android, but this work-around allows the Android build to run.

I also updated the build script for running the iosApp project.


liujoshua (approved):
LGTM. Retrying github action fixed the build

PR #42, opened 2021-07-09
Notifications handing

syoung-smallwisdom:
Also ended up fixing some issues with the timeline parsing and added unit tests using a test schedule that I had set up where I found the issues.


nategbrown9 (approved):
Looks good.

PR #44, opened 2021-07-13
Add support for showing history

syoung-smallwisdom:


Erin-Mounts (approved):
looks good to me

nategbrown9 (approved):
Looks good. We have some duplicate code in the UI layer for both iOS and Android. I'm going to look at adding a function on top of getPastSessions to do the extraction and sorting work.

PR #46, opened 2021-07-16
Add study repository

nategbrown9:
Still need to test, but here it is.


syoung-smallwisdom (approved):

syoung-smallwisdom (changes requested):
Upon further consideration, the flow doesn't really work correctly for the `StudyInfo` case. Really, that should not ever go to cache and should instead act more like sign in, maybe be included in the AuthManager since the goal is to use it to sign in the participant and then it isn't ever used within the app. Let's discuss when you have a chance. Thanks!

nategbrown9:
> Upon further consideration, the flow doesn't really work correctly for the `StudyInfo` case. Really, that should not ever go to cache and should instead act more like sign in, maybe be included in the AuthManager since the goal is to use it to sign in the participant and then it isn't ever used within the app. Let's discuss when you have a chance. Thanks!

So something like this on either AuthenticationRepo or StudyRepo? 


`public suspend fun getStudyInfo(studyId: String): ResourceResult<StudyInfo> {
        try {
            val studyInfo = studyApi.getStudyInfo(bridgeConfig.appId, studyId)
            return ResourceResult.Success(studyInfo, ResourceStatus.SUCCESS)
        } catch (err: Throwable) {
            println(err)
        }
        return ResourceResult.Failed(ResourceStatus.FAILED)
    }`

syoung-smallwisdom:
Yeah, was just wondering which made more sense to you?

syoung-smallwisdom (approved):

PR #47, opened 2021-07-21
Add NativeStudyManager and support for fetching study info

syoung-smallwisdom:


SwiftPackage/Sources/BridgeClientUI/Extensions/KotlinBase+ClientData.swift
```diff
@@ -35,27 +35,40 @@ import Foundation
 import BridgeClient
 import JsonModel
 
-extension AppConfig {
```
Erin-Mounts, line 38: update the filename in the comment block above (GitHub won't let me put this comment up there)

Erin-Mounts (approved):
one nit, otherwise looks good to me

PR #48, opened 2021-07-27
History record

nategbrown9:


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -545,4 +556,18 @@ data class ScheduledNotification(
...
+data class AssessmentHistoryRecord (
+    val instanceGuid: String,
+    val assessmentInfo: AssessmentInfo,
+    val startedOn: Instant,
+    val finishedOn: Instant,
+    val clientTimeZone: String? = null,
+    val clientData: kotlinx.serialization.json.JsonElement? = null) {
+
+    //TODO: Switch to using Instant math once we migrate to Kotlin 1.5 -nbrown 07/27/21
+    val minutes: Int = ((finishedOn.epochSeconds - startedOn.epochSeconds) / 60).toInt()
```
syoung-smallwisdom, line 570: `toInt()` will floor the value, when really we want to round it.

syoung-smallwisdom (commented):
Looks good. One comment about

syoung-smallwisdom (approved):
One comment about rounding. Otherwise, looks good

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
nategbrown9: The division operation is already doing a floor, `toInt()` is just converting the long into an int. In kotlin `long / long = long`. As soon as we move to kotlin 1.5.x (which I would like to do before our October release), we can use `val minutes = finishedOn.minus(startedOn).inMinutes.roundToInt()`

syoung-smallwisdom: Ah, didn't realize epochSeconds was a Long. (In iOS it's a Double)

PR #49, opened 2021-07-30
Build package

syoung-smallwisdom:
Changed iosApp to use the Swift package rather than building into this app. 

Note: I also moved the pre-build schema to only launch manually by running:

```
./xcframework_build_debug.sh
```

OR 

```
./xcframework_build_release.sh
```

from the command line. This is b/c I couldn't figure out a clean, safe way to switch between release and debug automatically since the xcframework path cannot change with debug. (bit annoying, that) Eventually, it would be nice to automate this, but that's a task for another day.


SwiftPackage/Binaries/BridgeClient.xcframework/ios-arm64/BridgeClient.framework/Info.plist
```diff
@@ -5,7 +5,7 @@
     <key>CFBundleExecutable</key>
     <string>BridgeClient</string>
     <key>CFBundleIdentifier</key>
-    <string>org.sagebionetworks.bridge.BridgeClient</string>
+    <string>BridgeClient</string>
```
Erin-Mounts, line 8: curious as to the reason behind this change

Erin-Mounts (approved):
one minor question, but looks good

SwiftPackage/Binaries/BridgeClient.xcframework/ios-arm64/BridgeClient.framework/Info.plist
syoung-smallwisdom: Good catch. I am not sure why it changed, but it's something in the Kotlin code that builds the framework. A 30 minute googling did not turn up anything. :(

PR #53, opened 2021-08-19
Added scroll buttons to the scrollview for the Today and History views

syoung-smallwisdom:
I have some concerns that for the history view with several pages this could seriously affect scrolling performance. I don't have enough experience with SwiftUI and `LazyVStack` to be able to say with confidence. In other words, I *think* using UUID's (instead of Int) as the page identifiers will allow for rendering only those views that are in the scroll window (even though height is calculated on the fly) but I'm not sure about paging for longer scroll windows.

### DIFF
--- a/SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
+++ b/SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
@@ -160,8 +160,8 @@ public struct CustomScrollView<Content : View>: View {
             // Figure out how many pages are needed.
             let pageCount = Int(ceil(scrollHeight/viewHeight))
             if pages.count > 0 {
-                // Lop off the end. If there are more pages that there are currently
-                // then doing do will refresh the last page with a new height by
+                // Lop off the end. If there are more pages than there are currently
+                // then doing so will refresh the last page with a new height by
                 // replacing the UUID that is used as the identifier.
                 pages = Array(pages[0..<(min(pageCount, pages.count) - 1)])
             }
--- a/SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
+++ b/SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
@@ -1,0 +1,250 @@
+//
+//  CustomScrollView.swift
+//
+//  Copyright © 2021 Sage Bionetworks. All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification,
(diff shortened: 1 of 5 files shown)


SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
```diff
@@ -0,0 +1,250 @@
...
+            guard scrollHeight > 0, viewHeight > 0, scrollEnabled
+            else {
+                // Do not load any pages if the view is longer than the content.
+                self.pages = []
+                return
+            }
+            // Figure out how many pages are needed.
+            let pageCount = Int(ceil(scrollHeight/viewHeight))
+            if pages.count > 0 {
+                // Lop off the end. If there are more pages that there are currently
```
Erin-Mounts: I'm having trouble parsing this second sentence--did you mean "than" rather than "that" and "so" rather than "do"?

Erin-Mounts (approved):
One question about a comment, otherwise looks good

SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
syoung-smallwisdom:
```suggestion
                // Lop off the end. If there are more pages than there are currently
```

SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
```diff
@@ -0,0 +1,250 @@
...
+                // then doing do will refresh the last page with a new height by
```
syoung-smallwisdom:
```suggestion
                // then doing so will refresh the last page with a new height by
```

SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomScrollView.swift
syoung-smallwisdom: Urgh, yes, that's what I meant. So... SwiftUI decides what needs to be refreshed based on the `id` of the model that is backing the view. In this case, the only way I could figure out how to add a "paging-style" of scrolling to a generic view was to add hidden views behind the main content. However, the last "page" has to be a non-standard height that will change when either the `viewHeight` or the `scrollHeight` change so I needed a way of retaining the ids for the other pages (so they don't *all* get refreshed) and while still replacing that last page (so it *does* get refreshed). I tried to add comments to explain this b/c it's very screwy but apparently I can't type. Oops. And thanks!

PR #54, opened 2021-08-30
Bmc 215/upload and encrypt

syoung-smallwisdom:
So... I am not sure why you weren't getting the call to uploadAndEncrypt but I did find some bugs while testing.


Erin-Mounts (approved):
Looks good. 

FWIW I tried to chase the call chain back to see where it *should* be getting called from and at some point the trail just ended, so then I tried doing measures with breakpoints set and they never got hit (created myself a participant in the demo study).

PR #57, opened 2021-09-02
Add unit tests for the today timeline

syoung-smallwisdom:
Not actually a lot to unit test. Basically, this is mostly stuff in the UI layer though there is a little bit of date calculation.


nategbrown9 (approved):
Looks good.

PR #61, opened 2021-09-10
Include studyId in ScheduledAssessmentReference

liujoshua:


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -531,6 +532,7 @@ data class ScheduledSessionWindow (
 }
 
 data class ScheduledAssessmentReference (
+    val studyId: String,
```
syoung-smallwisdom: Complete and total nit: Can you move this to after the `instanceGuid`. That's the unique identifier and I like listing it first. :) Thanks!

syoung-smallwisdom: Also, this is for archiving, right? (Only reason I can think of why you need the studyId) There are other bits that you will need to add to the metadata or to find this assessment instance.

For that, on iOS I ended up with a model that pulls all the stuff that goes into the archive metadata into a single model object. (Note: I also have to make it a struct b/c kotlin objects will crash if accessed off the main thread)

```
/// Thread-safe assessment schedule info. Kotlin objects can only be accessed from the main thread or it causes
/// a crash so this must be initialized on the main thread.
public struct AssessmentScheduleInfo {
    
    /// The ``BridgeClient.ScheduledAssessment/instanceGuid`` for this scheduled assessment.
    public let instanceGuid: String
    
    /// Information about the session.
    public let session: Session
    
    /// Information about the assessment.
    public let assessmentInfo: Info
    
    public init(session: NativeScheduledSessionWindow, assessment: NativeScheduledAssessment) {
        self.instanceGuid = assessment.instanceGuid
        self.session = Session(instanceGuid: session.instanceGuid,
                               eventTimestamp: session.eventTimestamp,
                               scheduledOn: session.startDateTime)
        self.assessmentInfo = Info(identifier: assessment.assessmentInfo.identifier!,
                                   key: assessment.assessmentInfo.key,
                                   guid: assessment.assessmentInfo.identifier,
                                   label: assessment.assessmentInfo.label)
    }
    
    /// A thread-safe struct with information required to load and save the results from an assessment.
    public struct Info : Hashable, Codable {
        /// The ``BridgeClient.AssessmentInfo/identifier``.
        public let identifier: String
        /// The ``BridgeClient.AssessmentInfo/key``.
        public let key: String?
        /// The ``BridgeClient.AssessmentInfo/guid``.
        public let guid: String?
        /// The ``BridgeClient.AssessmentInfo/label``.
        public let label: String?
    }
    
    /// A thread-safe struct with information needed to find and update the Kotlin ``BridgeClient.NativeScheduledSessionWindow``
    /// for the session to which this assessment belongs.
    public struct Session : Hashable, Codable {
        /// The ``BridgeClient.ScheduledSession/instanceGuid`` for this scheduled assessment.
        public let instanceGuid: String
        /// The ``BridgeClient.StudyActivityEvent/timestamp`` for the event used as the reference
        /// when *this* schedule was built. This value is mutable on the server so the timestamp is used rather
        /// than the event id.
        public let eventTimestamp: String
        /// The calculated start time for when this session opens.
        public let scheduledOn: Date
    }
}
```

liujoshua: That's a good point, I'll change the order.

I want to add studyId to retrieve recorder configuration. We use the ScheduledAssessment to launch tasks and keeping scheduled assessment with its study lets the launcher retrieve the info it needs from AppConfig and Study clientData. How did you did this on iOS?

liujoshua: As far as I can tell, Mobile Toolbox only lets you sign into one study at a time, but the server APIs let you have multiple studies in your UserSession and iirc multiple studies in a single app is on our roadmap.

syoung-smallwisdom: Ah, right. On iOS I have to use wrappers and environment variables b/c of some limitations on using Kotlin objects so the thing that starts the assessment already has the Study object.

ie. I can't access the repo directly.

syoung-smallwisdom: WRT Mobile Toolbox vs. other apps. The model supported by the server is intended to allow for multiple studies in a single app for cases like mPower where someone might join more than one study and have schedules associated with both. B/c of how SwiftUI works and some limitations of Kotlin Multiplatform, I have to wrap the objects from the server. I end up with what is effectively a singleton of a manager that is tailored to the single-study sign in. 

WRT the background recorders, the reason all that stuff is in ConfigElement and clientData is b/c we lack confidence that this is really a design that will stick. There are too many edge cases that the researchers didn't consider when making the original promise to support them. So for now, they are *only* applicable to the single-study Mobile Toolbox App. Should this become a long-term product that we need to support, there will need to be some changes to better accommodate wrapping a generic assessment while not clobbering the listeners that run on the assessment. (For example, microphone on the ones that rely upon the speaker.)

syoung-smallwisdom (approved):
Whoops, sorry. Forgot to approve. Looks good

PR #62, opened 2021-09-10
File uploads for iOS

Erin-Mounts:


syoung-smallwisdom (commented):
Nit: Can you rename the git comments to *not* be "wip"? Thanks!

Erin-Mounts:
> Nit: Can you rename the git comments to _not_ be "wip"? Thanks!

sure, how do I do that?

Erin-Mounts:
> > Nit: Can you rename the git comments to _not_ be "wip"? Thanks!
> 
> sure, how do I do that?

never mind, I googled it

.idea/jarRepositories.xml
```diff
@@ -21,25 +21,35 @@
...
+    </remote-repository>
+    <remote-repository>
+      <option name="id" value="maven5" />
+      <option name="name" value="maven5" />
+      <option name="url" value="https://dl.bintray.com/sage-bionetworks/BridgeClientKMM" />
+    </remote-repository>
+    <remote-repository>
+      <option name="id" value="maven4" />
+      <option name="name" value="maven4" />
+      <option name="url" value="https://oss.sonatype.org/content/repositories/snapshots" />
```
syoung-smallwisdom: Um... ?

.idea/runConfigurations.xml
```diff
@@ -3,7 +3,6 @@
   <component name="RunConfigurationProducerService">
     <option name="ignoredProducers">
       <set>
-        <option value="com.android.tools.idea.compose.preview.runconfiguration.ComposePreviewRunConfigurationProducer" />
```
syoung-smallwisdom, line 6: ?

SwiftPackage/Sources/BridgeClientUI/BridgeClientAppManager.swift
```diff
@@ -156,10 +166,12 @@ open class BridgeClientAppManager : ObservableObject {
         DispatchQueue.global().async {
 
             // Encrypt the files.
+            var encryptedArchives = [String : URL]()
             if let path = self.pemPath {
                 archives.forEach { archive in
                     do {
-                        try archive.encryptArchive(using: path)
+                        let encryptedArchive = try archive.encryptArchive(using: path)
+                        encryptedArchives[archive.identifier] = encryptedArchive
```
syoung-smallwisdom: FYI: This is the archiving code:

```
    /// Encrypt the archive using the given pem file.
    /// - note: This will also delete the unencrypted zip file as a part of cleanup.
    /// - parameters:
    ///     - pemPath: Path to the .pem file to use with CMS encryption.
    ///     - url: Optional URL for the encrypted file. If null, then the  `archiveURL` will be appended with ".encrypted".
    @discardableResult
    public func encryptArchive(using pemPath: String, to url: URL? = nil) throws -> URL {
        guard let _ = self.archive, let archiveURL = self.archiveURL else {
            throw BridgeArchiverError.archiveClosed
        }
        let encryptedURL = url ?? archiveURL.appendingPathExtension("encrypted")
        let data = try Data(contentsOf: archiveURL)
        let encryptedData = try CMSSupport.cmsEncrypt(data, identityPath: pemPath)
        try encryptedData.write(to: encryptedURL)
        try FileManager.default.removeItem(at: archiveURL)
        self.encryptedURL = encryptedURL
        self.archive = nil
        self.archiveURL = nil
        return encryptedURL
    }
```

So below you could do this:

```
                archives.forEach { archive in
                    guard let url = archive.encryptedURL else {
                        print("WARNING! Cannot upload \(archive.identifier)")
                        return
                    }
                    let id = archive.identifier
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+import UIKit
```
syoung-smallwisdom: What uses UIKit?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+class BridgeJsonParser {
+    /// The .iso8601 date encoding/decoding strategy for JSONEncoder/Decoder uses ISO8601DateFormatter,
+    /// which (a) is not a subclass of DateFormatter and (b) only supports the RFC 3339 profile, which does not include
+    /// fractional seconds. Bridge always includes fractional seconds in dates.
+    
+    static let iso8601FormatterWithFractionalSeconds: DateFormatter = {
+        let formatter = DateFormatter()
+        formatter.calendar = Calendar(identifier: .iso8601)
+        formatter.locale = Locale(identifier: "en_US_POSIX")
+        formatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSXXXXX"
```
syoung-smallwisdom: What is the difference between this and `JsonModel.ISO8601TimestampFormatter` which uses "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ" and is used elsewhere for bridge timestamps?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .formatted(BridgeJsonParser.iso8601FormatterWithFractionalSeconds)
+        return decoder
+    }()
+}
+
+class BackgroundNetworkManager: NSObject, URLSessionBackgroundDelegate {
+
+    /// A singleton instance of the manager.
+    public static let shared = BackgroundNetworkManager()
```
syoung-smallwisdom: You have this marked as public but the class itself is internal. Why?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+
+    /// The identifier (base) for the background URLSession. App extensions will append a unique string to this to distinguish
+    /// their background sessions.
+    // TODO: Allow users of this singleton to specify their own individually unique identifiers.
+    public static let backgroundSessionIdentifier = "org.sagebase.backgroundnetworkmanager.session"
```
syoung-smallwisdom: You have this marked as public but the class itself is internal. Why?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        return delegateQueue
+    }()
+            
+    /// The singleton background URLSession.
+    public static var _backgroundSession: URLSession? = nil
+    
+    /// A singleton map of pending background URLSession completion handlers that have been passed in from the app delegate and not yet called.
+    public static var backgroundSessionCompletionHandlers = [String : () -> Void]()
+    
+    private static func isRunningInAppExtension() -> Bool {
```
syoung-smallwisdom: It's a singleton, why static?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+    
+    /// For decoding objects received from Bridge.
+    static let jsonDecoder: JSONDecoder = {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .formatted(BridgeJsonParser.iso8601FormatterWithFractionalSeconds)
+        return decoder
+    }()
+}
+
+class BackgroundNetworkManager: NSObject, URLSessionBackgroundDelegate {
```
syoung-smallwisdom, line 61: You're mixing internal, public, and private and using static on a singleton. Kinda confusing to read.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+    ///  Created lazily.
+    // TODO: When supporting multiple sessions (below), does it make sense to share one delegate queue among all of them?
+    public static let sessionDelegateQueue: OperationQueue = {
+        let delegateQueue = OperationQueue()
+        delegateQueue.maxConcurrentOperationCount = 1
+        return delegateQueue
+    }()
+            
+    /// The singleton background URLSession.
+    public static var _backgroundSession: URLSession? = nil
```
syoung-smallwisdom: a public ivar shouldn't have an `_` underscore.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        
+        let baseUrl = bridgeBaseURL()
+        guard let bridgeUrl = URL(string: urlString, relativeTo: baseUrl) else {
+            fatalError("Unable to create URL object from string '\(urlString)' relative to \(baseUrl)")
+        }
+        
+        return bridgeUrl
+    }
+    
+    func queryString(from parameters: Encodable?) -> String? {
```
syoung-smallwisdom: Why are you passing in the parameters as an Encodable? Why not just pass in `[String : String]`?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+
+    /// We use this custom HTTP request header as a hack to keep track of how many times we've retried a request.
+    let retryCountHeader = "X-SageBridge-Retry"
+    
+    /// This sets the maximum number of times we will retry a request before giving up.
+    let maxRetries = 5
+    
+    /// If set, URLSession(Data/Download)Delegate method calls received by the BackgroundNetworkManager
+    /// will be passed through to this object for further handling.
+    // TODO: Allow multiple delegates, one for each unique background session identifier.
```
syoung-smallwisdom: sign and date all TODO comments.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+    public var backgroundTransferDelegate: URLSessionBackgroundDelegate?
+    
+    /// The queue used for calling background session delegate methods.
+    ///  Also used for creating the singleton background session itself in a thread-safe way.
+    ///  Created lazily.
+    // TODO: When supporting multiple sessions (below), does it make sense to share one delegate queue among all of them?
```
syoung-smallwisdom: sign and date all TODO comments.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        let delegateQueue = OperationQueue()
+        delegateQueue.maxConcurrentOperationCount = 1
+        return delegateQueue
+    }()
+            
+    /// The singleton background URLSession.
+    public static var _backgroundSession: URLSession? = nil
+    
+    /// A singleton map of pending background URLSession completion handlers that have been passed in from the app delegate and not yet called.
+    public static var backgroundSessionCompletionHandlers = [String : () -> Void]()
```
syoung-smallwisdom: public?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        task.taskDescription = taskDescription
+        task.resume()
+        return task
+    }
+    
+    public func restore(backgroundSession: String, completionHandler: @escaping () -> Void) {
+        if backgroundSession.starts(with: BackgroundNetworkManager.backgroundSessionIdentifier) {
+            BackgroundNetworkManager.backgroundSessionCompletionHandlers[backgroundSession] = completionHandler
+            BackgroundNetworkManager.sessionDelegateQueue.addOperation {
+                // TODO: Update this when support is added for multiple named background sessions.
```
syoung-smallwisdom: sign and date TODO's

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+                    // instances of the same extension, if such a thing is possible, wouldn't interfere with
+                    // each other.
+                    sessionIdentifier.append(UUID().uuidString)
+                }
+                
+                self.createBackgroundSession(with: sessionIdentifier)
+            }
+            BackgroundNetworkManager.sessionDelegateQueue.addOperations([createSessionOperation], waitUntilFinished: true)
+        }
+        return BackgroundNetworkManager._backgroundSession!
```
syoung-smallwisdom, line 150: This code feels pretty brittle. I'm struggling to understand why you aren't using dispatch?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+            } catch let error {
+                debugPrint("Error trying to jsonEncode value:\n\(value)\n\(error)")
+                continue
+            }
+            guard let encodedValueString = String(data: encodedValue, encoding: .utf8)?.addingPercentEncoding(withAllowedCharacters: allowedChars) else { continue }
+            guard let encodedParam = param.addingPercentEncoding(withAllowedCharacters: allowedChars) else { continue }
+            
+            queryParams.append("\(encodedParam)=\(encodedValueString)")
+        }
+        return queryParams.joined(separator: "&")
```
syoung-smallwisdom: So, the json encoder is just taking the parameters and wrapping the value in quotation marks and encoding as UTF8. You then unencoded as UTF8 and percent encode it. That's kinda confusing.

Instead you could use reduce:
```
func queryString(from parameters: [String : String]?) -> String? {
    parameters.reduce("") {
        guard let encodedParam = $1.key.addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed),
              let encodedValueString = "\"\($1.value)\"".addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed)
        else {
            return $0
        }
        
        let next = "\(encodedParam)=\(encodedValueString)"
        return $0.isEmpty ? next : "\($0)&\(next)"
    }
}
```

Or use compact map:
```
func queryString(from parameters: [String : String]?) -> String? {
    parameters.compactMap { (param, value) in
        guard let encodedParam = param.addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed),
              let encodedValueString = "\"\(value)\"".addingPercentEncoding(withAllowedCharacters: CharacterSet.urlQueryAllowed)
        else {
            return nil
        }
        return "\(encodedParam)=\(encodedValueString)"
    }.joined(separator: "&")
}
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        var request = URLRequest(url: bridgeURL(for: URLString))
+        request.httpMethod = method
+        request.httpShouldHandleCookies = false
+        addBasicHeaders(to: &request)
+        
+        if let headers = headers {
+            for (header, value) in headers {
+                request.addValue(value, forHTTPHeaderField: header)
+            }
+        }
```
syoung-smallwisdom: So... using an `inout` is a little weird for Swift. Would be better to do this:

```
extension URLRequest {
    mutating fileprivate func addHeaders(_ headers: [String : String]?) {
        let userAgentHeader = IOSBridgeConfig().userAgent
        self.setValue(userAgentHeader, forHTTPHeaderField: "User-Agent")

        let acceptLanguageHeader = Locale.preferredLanguages.joined(separator: ", ")
        self.setValue(acceptLanguageHeader, forHTTPHeaderField: "Accept-Language")

        self.setValue("no-cache", forHTTPHeaderField: "cache-control")
        
        headers?.forEach {
            self.addValue($0.key, forHTTPHeaderField: $0.value)
        }
    }
}
```

And then in the function that builds the request call
```
request.addHeaders(headers)
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        return request
+    }
+    
+    func request<T>(method: String, URLString: String, headers: [String : String]?, parameters: T) -> URLRequest where T: Encodable {
+        var URLString = URLString
+        let isGet = (method.uppercased() == "GET")
+        
+        // for GET requests, the parameters go in the query part of the URL
+        if isGet {
+            if let queryString = queryString(from: parameters), !queryString.isEmpty {
```
syoung-smallwisdom:
```
if isGet, let queryString = queryString(from: parameters), !queryString.isEmpty {
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        
+        return request
+    }
+    
+    public func downloadFile(from URLString: String, method: String, httpHeaders: [String : String]?, taskDescription: String) -> URLSessionDownloadTask {
+        let request = self.request(method: method, URLString: URLString, headers: httpHeaders)
+        return self.downloadFile(with: request, taskDescription: taskDescription)
+    }
+    
+    public func downloadFile<T>(from URLString: String, method: String, httpHeaders: [String : String]?, parameters: T?, taskDescription: String) -> URLSessionDownloadTask where T: Encodable {
```
syoung-smallwisdom: If your `method` has to be either GET or POST, then it would be less brittle to use an enum.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        request.allHTTPHeaderFields = httpHeaders
+        request.httpMethod = "PUT"
+        let task = backgroundSession().uploadTask(with: request, fromFile: fileURL)
+        task.taskDescription = taskDescription
+        task.resume()
+        return task
+    }
+    
+    public func restore(backgroundSession: String, completionHandler: @escaping () -> Void) {
+        if backgroundSession.starts(with: BackgroundNetworkManager.backgroundSessionIdentifier) {
```
syoung-smallwisdom: Use a guard and exit early.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -0,0 +1,467 @@
...
+        let resumeTask = session.downloadTask(withResumeData: resumeData)
+        resumeTask.taskDescription = task.taskDescription
+        resumeTask.resume()
+    }
+    
+    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
+        if let nsError = error as NSError? {
+            // if there's resume data from a download task, use it to retry
+            if let downloadTask = task as? URLSessionDownloadTask,
+               let resumeData = nsError.userInfo[NSURLSessionDownloadTaskResumeData] as? Data {
```
syoung-smallwisdom: Why nested if statement?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+    /// Decodes given Decodable type from given array or dictionary
+    func decode<T>(_ type: T.Type, from json: Any) throws -> T where T: Decodable {
+        // if it's already the requested, simple plist type, don't bother serializing/decoding it
+        if json is T && (json is NSData || json is NSString || json is NSDate || json is NSNumber) {
+            return json as! T
+        }
+        let jsonData = try JSONSerialization.data(withJSONObject: json, options: [])
+        return try jsonDecoder.decode(type, from: jsonData)
+    }
+}
```
syoung-smallwisdom, line 82: If it ain't broke... still, you are reinventing the wheel. See `JsonModel` - it has utilities for encoding/decoding.

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
```diff
@@ -0,0 +1,240 @@
...
+    public static let SBBParticipantFileUploadToS3Failed = Notification.Name(rawValue: "SBBParticipantFileUploadToS3Failed")
+}
+
+class ParticipantFileUploadAPI: BridgeFileUploadAPITyped {
+    typealias TrackingType = ParticipantFile
+    typealias UploadRequestType = ParticipantFile
+    typealias UploadRequestResponseType = ParticipantFile
+    
+    /// A singleton instance of the API.
+    public static let shared = ParticipantFileUploadAPI()
```
syoung-smallwisdom, line 66: Again, the class is internal. Why the public marked accessors?

SwiftPackage/Sources/BridgeClientUI/Models/PlatformConfigImpl.swift
```diff
@@ -33,26 +33,30 @@
...
-
+public class PlatformConfigImpl : IOSPlatformConfig {
+        
+    public let appGroupIdentifier: String?
     public let appId: String
     public let appName: String
     public let appVersionName: String
     public let appVersion: Int32
     public let deviceName: String
+    public let bridgeEnv: PlatformConfigBridgeEnv
```
syoung-smallwisdom: `bridgeEnvironment`

bridgeClient/.idea/workspace.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
syoung-smallwisdom: Do we know why this file was added?

syoung-smallwisdom (commented):
Needs to point at release build.

I got tired and didn't review the really long file. I'll keep later after taking a break.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+import UIKit
```
syoung-smallwisdom, line 35: Again, not sure why this uses UIKit?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+    func extendedAttribute(forName name: String) throws -> Data  {
+
+        let data = try self.withUnsafeFileSystemRepresentation { fileSystemPath -> Data in
+
+            // Determine attribute size:
+            let length = getxattr(fileSystemPath, name, nil, 0, 0, 0)
+            guard length >= 0 else { throw URL.posixError(errno) }
+
+            // Create buffer with required size:
+            var data = Data(count: length)
```
syoung-smallwisdom, line 98: Bit messy to use the same ivar name both here and above. Maybe do this instead?

```
try self.withUnsafeFileSystemRepresentation { fileSystemPath -> Data in
```

and leave off the `let data = ` and `return data`

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+        try self.withUnsafeFileSystemRepresentation { fileSystemPath in
+            let result = removexattr(fileSystemPath, name, 0)
+            guard result >= 0 else { throw URL.posixError(errno) }
+        }
+    }
+
+    /// Get list of all extended attributes.
+    func listExtendedAttributes() throws -> [String] {
+
+        let list = try self.withUnsafeFileSystemRepresentation { fileSystemPath -> [String] in
```
syoung-smallwisdom: again, the nested ivars with the same name is a bit messy.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            
+            // Set its state as notifyingBridge
+            uploadManager.persistMapping(from: relativePath, to: uploadMetadata as! BridgeFileUploadMetadata<TrackingType>, defaultsKey: uploadManager.notifyingBridgeUploadSucceededKey)
+
+            let headers = [
+                "Bridge-Session" : sessionToken
+            ]
+
+            let _ = uploadManager.netManager.downloadFile(from: notifyUrl, method: "POST", httpHeaders: headers, taskDescription: relativePath)
+            }
```
syoung-smallwisdom:
```suggestion
        }
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            if !self.isUploadSessionExpired(for: uploadMetadata) {
+                // if we have an upload URL and it's not expired, just retry upload to S3
+                // -- restore tracking of what file it's a copy of
+                self.uploadManager.persistMapping(from: relativePath, to: fileUrl.path, defaultsKey: self.uploadManager.bridgeFileUploadsKey)
+                // -- retry upload to S3
+                self.uploadManager.uploadToS3(uploadApi: self, bridgeUploadMetadata: uploadMetadata, invariantFilePath: relativePath)
+            }
+            else {
+                // request a fresh upload URL from Bridge
+                let _ = self.uploadManager.requestUploadURL(uploadApi: self, invariantFilePath: relativePath, fileUrl: fileUrl, uploadMetadata: uploadMetadata)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            }
+            
+            // Set its state as uploadRequested
+            uploadManager.persistMapping(from: relativePath, to: uploadMetadata as! BridgeFileUploadMetadata<TrackingType>, defaultsKey: uploadManager.uploadURLsRequestedKey)
+
+            let headers = [
+                "Bridge-Session" : sessionToken
+            ]
+
+            let _ = uploadManager.netManager.downloadFile(from: requestString, method: "POST", httpHeaders: headers, parameters: bridgeUploadObject, taskDescription: relativePath)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            }
+            
+            // Set its state as notifyingBridge
+            uploadManager.persistMapping(from: relativePath, to: uploadMetadata as! BridgeFileUploadMetadata<TrackingType>, defaultsKey: uploadManager.notifyingBridgeUploadSucceededKey)
+
+            let headers = [
+                "Bridge-Session" : sessionToken
+            ]
+
+            let _ = uploadManager.netManager.downloadFile(from: notifyUrl, method: "POST", httpHeaders: headers, taskDescription: relativePath)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            // now just bail--we'll handle it if and when it makes it to `urlSession(_:, task:, didCompleteWithError:)`.
+            return
+        }
+
+        // remove the participant file from the uploadRequested list, retrieving its
+        // associated upload metadata; or, if not found there, check for and remove it from
+        // the notifyingBridge list and we're done
+        
+        guard let uploadMetadata = uploadApi.fetchUploadRequested(for: invariantFilePath) else {
+            guard let _ = uploadApi.fetchNotifyingBridge(for: invariantFilePath) else {
```
syoung-smallwisdom, line 876: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+                // if it wasn't in either list, log it and bail
+                debugPrint("Unexpected: Unable to retrieve upload metadata for \(invariantFilePath) from either upload request map or notifying Bridge map.")
+                return
+            }
+
+            // since it was notifying bridge, we're all done at this point, so we
+            // can now tie up any loose ends:
+            
+            // -- remove from retry queue, in case it somehow ended up there
+            let _ = uploadApi.dequeueForRetry(relativePath: invariantFilePath)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            
+            // -- remove the temp file -> original file mapping
+            let _ = removeMapping(String.self, from: invariantFilePath, defaultsKey: self.bridgeFileUploadsKey)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+        // ...and make a URL from it
+        let fileUrl = URL(fileURLWithPath: filePath)
+        
+        // add the file to the uploadingToS3 list
+        uploadApi.markUploadingToS3(for: invariantFilePath, uploadMetadata: bridgeUploadMetadata)
+
+        
+        // upload the file to S3
+        let headers = uploadApi.s3UploadHeaders(for: bridgeUploadMetadata)
+        let _ = self.netManager.uploadFile(fileUrl, httpHeaders: headers, to: uploadUrl, taskDescription: invariantFilePath)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+        // remove the file from the temp -> orig mappings, and retrieve the original file path
+        guard let originalFilePath = removeMapping(String.self, from: invariantFilePath, defaultsKey: self.bridgeFileUploadsKey) else {
+            debugPrint("Unexpected: No original file path found mapped from temp file path \(invariantFilePath)")
+            return
+        }
+        
+        // remove the file from the retry queue if it's in there (as it might be if an URLSession was invalidated
+        // due to an error)
+        
+        let _ = uploadApi.dequeueForRetry(relativePath: invariantFilePath)
```
syoung-smallwisdom: Use `@discardableResult` to mark results as "ignorable" and then you don't need the `let _ =`. Alternatively, please add a comment to explain *why* you are ignoring the result.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+    /// Session delegate method.
+    public func urlSession(_ session: URLSession, didBecomeInvalidWithError error: Error?) {
+        if error == nil {
+            // If the URLSession was deliberately invalidated (i.e., error is nil) then we assume
+            // the intention is to cancel and forget all incomplete uploads, including retries.
+            // This might be used e.g. to clear out all pending uploads if we receive a 412
+            // (Not Consented) error from Bridge or if the participant withdraws from the study.
+            self.reset()
+        }
+        else {
```
syoung-smallwisdom:
```
guard error == nil 
else {
    // If the URLSession was deliberately invalidated (i.e., error is nil) then we assume
    // the intention is to cancel and forget all incomplete uploads, including retries.
    // This might be used e.g. to clear out all pending uploads if we receive a 412
    // (Not Consented) error from Bridge or if the participant withdraws from the study.
    self.reset()
    return
}
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1292 @@
...
+            self.uploadQueue.addOperation(block)
+        }
+    }
+    
+    /// MARK: Handle converting between sandbox-relative and fully-qualified file paths.
+    private lazy var baseDirectory: String = {
+        var baseDirectory: String
+        if let appGroupIdentifier = IOSBridgeConfig().appGroupIdentifier, appGroupIdentifier.count > 0 {
+            let baseDirUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier)
```
syoung-smallwisdom:
```
if let appGroupIdentifier = IOSBridgeConfig().appGroupIdentifier, appGroupIdentifier.count > 0,
             let baseDirUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: appGroupIdentifier) {
```

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
```diff
@@ -0,0 +1,240 @@
...
+// FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+// SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+// CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+//import Foundation
+import UIKit
```
syoung-smallwisdom: Again, why UIKit?

Ideally, it would be nice to be able to run this against macOS for GitHub Action swift testing. Swift PM doesn't currently allow setting which os versions a library is built for.

syoung-smallwisdom (changes requested):
Lots of nits that don't look like they are breaking changes but might make things a little more future-proof. 

I'm not approving just yet b/c I am concerned about all the Android Studios changes. I don't understand the build system well enough to know if these break things.

.idea/jarRepositories.xml
Erin-Mounts: No idea. I just clicked the Sync Project with Gradle Files button in Android Studio, and ran your build script. I didn't touch this file myself.

.idea/runConfigurations.xml
Erin-Mounts: ¯\\_(ツ)\_/¯

SwiftPackage/Sources/BridgeClientUI/BridgeClientAppManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Apparently nothing? I think this was part of the boilerplate when I created the blank file and I forgot to remove it.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Apparently nothing. I just didn't know where to look for that when I was making this. Thanks!

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Habit. Or possible copied from a public class without thinking.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Because in previous implementations, the AppDelegate might need to know this to properly handle background url session events. But since the code that handles this now lives in this same module, and the AppDelegate just calls that, it doesn't need to be.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: I wasn't initially sure this would be a singleton class, and never updated it once it was. I think the BridgeSDK implementation also was static, which made it available even if you were to register a custom network manager. But that doesn't apply here.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: yeah, that's how it evolved as I developed it. sorry about that.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Because the (only) function that calls this one works for all types of requests, not just GETs, and for POSTs the parameters can be passed in as e.g. an UploadRequest object or a ParticipantFile object. That func just passes the parameters arg straight through to this one. I suppose I could have done the check in the caller, but this is ported directly from the original Objective-C implementation and I honestly don't remember my reasoning (if any) for doing it here instead.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: or get rid of them when they no longer apply ;)

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: deleted

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: yeah, yeah :P

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: deleted

.idea/jarRepositories.xml
nategbrown9: All the remote repositories being added here are no longer used. Not sure if this would break anything, but probably safer to just revert the changes to this file.

.idea/runConfigurations.xml
nategbrown9: Don't think this should break anything.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: The code in BridgeSDK used dispatch_once to accomplish this, but that's not available in Swift. Or do you mean why use an OperationQueue instead of a DispatchQueue? I guess mainly because I was already using them since the background URLSession delegate queue is an OperationQueue, and it seemed weirder to me to mix-and-match the two types of concurrency management queues.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
syoung-smallwisdom: Ock, just did a "What!?! No... That can't be right" and then googled it. Huh. Sorry.

Anyhow, good to know! I guess maybe we revisit in a month when Xcode 13 is released. ¯\_(ツ)_/¯

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Yep. Again, this was ported directly from Objective-C. Fixed.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Because if there's an error that *isn't* coupled with resume data from a download task, then I want to fall through and (1) handle any HTTP errors and (2) pass the error along to the backgroundTransferDelegate (i.e. the BridgeFileUploadManager).

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
Erin-Mounts: Because this class is meant to be public so an app can actually use it. Fixing that.

SwiftPackage/Sources/BridgeClientUI/Models/PlatformConfigImpl.swift
Erin-Mounts: 👍 (was just echoing what was previously in BridgeSDK/BridgeInfo.plist)

bridgeClient/.idea/workspace.xml
Erin-Mounts: I sure don't, so if you don't either, then no, "we" don't.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: so that it can register an observer for the UIApplication.didBecomeActiveNotification to retry stuck uploads.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍 (except != instead of == so the else clause happens if it does equal nil)

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: I should have caught that when porting from Obj-C--thanks!

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
Erin-Mounts: Boilerplate. Deleted. 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
```diff
@@ -0,0 +1,239 @@
...
+// OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+//import Foundation
+import UniformTypeIdentifiers
+import CoreServices
+import BridgeClient
+
+public struct ParticipantFile: Codable {
```
syoung-smallwisdom: FYI: 

1. This is public with all the properties as internal so outside this framework they cannot be seen. 
2. The initializer is internal so it can only be initialized internally.
3. Any reason why the properties are all read/write?

SwiftPackage/Sources/BridgeClientUI/File Uploading/StudyDataUploadAPI.swift
```diff
@@ -0,0 +1,296 @@
...
+    
+    /// Notification name posted by the `BridgeFileUploadManager` when a study file upload attempt to S3 fails unrecoverably.
+    public static let SBBStudyFileUploadToS3Failed = Notification.Name(rawValue: "SBBStudyFileUploadToS3Failed")
+    
+    /// Notification name posted by the `BridgeFileUploadManager` when attempting to notify Bridge
+    /// of a successful upload fails unrecoverably.
+    public static let SBBStudyFileUploadBridgeNotificationFailed = Notification.Name(rawValue: "SBBStudyFileUploadBridgeNotificationFailed")
+}
+
+class StudyDataUploadAPI: BridgeFileUploadAPITyped {
```
syoung-smallwisdom: Should this also be public?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -0,0 +1,1295 @@
...
+    /// Let Bridge know the upload succeeded.
+    /// Meant for internal use only.
+    func notifyBridgeUploadSucceeded(relativePath: String, uploadMetadata: BridgeFileUploadMetadataBlob)
+}
+
+/// BridgeFileUploadAPITyped extends BridgeFileUploadAPI with an associated type, which is
+/// the Swift (Codable) type of the object the particular API uses to pass parameters back and forth.
+/// To support a new Bridge file upload API, create a class that implements this protocol to handle the peculiarities
+/// of that particular API, and register an instance of it with the singleton BridgeFileUploadManager for that API.
+protocol BridgeFileUploadAPITyped : BridgeFileUploadAPI {
```
syoung-smallwisdom, line 281: Should this protocol be public?

bridgeClient/.idea/workspace.xml
syoung-smallwisdom: Hmmm.... @nategbrown9 ?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
syoung-smallwisdom: Yes... I am not seeing why this wouldn't do the same:

```
        if let nsError = error as NSError?,
           let downloadTask = task as? URLSessionDownloadTask,
           let resumeData = nsError.userInfo[NSURLSessionDownloadTaskResumeData] as? Data {
            // if there's resume data from a download task, use it to retry
            self.retryFailedDownload(downloadTask, for: session, resumeData: resumeData)
            return
        }
```

syoung-smallwisdom (approved):
Looks good. Couple of follow up questions.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BackgroundNetworkManager.swift
Erin-Mounts: Ah. Once again, I'm translating Obj-C too literally and not into idiomatic Swift. Fixed.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: 👍

SwiftPackage/Sources/BridgeClientUI/File Uploading/ParticipantFileUploadAPI.swift
Erin-Mounts: 1. Good point, removing `public`. If this ever gets refactored to not require it to be Codable (or the Kotlin model objects can be Codable) then this should just use the Kotlin model object.
2. See 1.
3. Only the first three are required, but if I make them all `let` then I can't initialize one without specifying all properties. Fixed those three. I'd like to have made `type` also be `let`, but then it gives a warning that the field can't get overwritten during deserialization, which is fine but I wanted to remove the warning.

SwiftPackage/Sources/BridgeClientUI/File Uploading/StudyDataUploadAPI.swift
Erin-Mounts: oops yeah 😅

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: yeah, again oops

Erin-Mounts: hmm, on second thought--if I make this public, then I get cascading requirements to make a whole bunch of stuff public that I don't think really needs (or ought) to be. If we ever need to support a new Bridge upload API, that support should go in BridgeClient anyway so I will leave this as internal.

bridgeClient/.idea/workspace.xml
nategbrown9: It could contain personalized settings for Android Studio, so we should remove it and add it to git ignore. https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

Erin-Mounts: @nategbrown9 does this look good for adding to the end of the .gitignore (taken from that link) before I do the dance to remove newly-ignored files from the repo?

```
# Intellij IDE
#
# See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

.idea/workspace.xml
.idea/usage.statistics.xml
.idea/shelf/
```

Erin-Mounts: Took me a minute to figure out what's going on here--there's no .gitignore in bridgeClient and apparently the pattern matching in .gitignore files isn't recursive, at least when the pattern includes a partial path? ¯\\_(ツ)\_/¯ anyway, adding a .gitignore to that folder now...

Erin-Mounts: @nategbrown9 you might want to look over the relevant commit (17d229fae0a77b75904309846359e1c35cb51a09) since it ended up removing some of your user-specific Xcode files for iosApp...

PR #63, opened 2021-09-13
Advance firstInstant regardless of the intervalPeriod size.

nategbrown9:
Researcher UI supports  reminder intervals of minutes, hours, and days.


syoung-smallwisdom (approved):
looks good

PR #64, opened 2021-09-14
Fix android studios

syoung-smallwisdom:
This runs `check`, `build`, and the script for building the xcframework for the swift package (debug and release).

Note: I did *not* use the XCFramework() gradle build that Kotlin uses b/c that doesn't put it in the correct location for the swift package and I can't figure out how to copy it there - didn't seem worth the time to figure it out since I was already building it where I want it to go. ¯\_(ツ)_/¯

### DIFF
--- a/bridge-client/build.gradle.kts
+++ b/bridge-client/build.gradle.kts
@@ -1,5 +1,6 @@
 import org.jetbrains.kotlin.gradle.plugin.mpp.Framework
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.apple.XCFramework
 
 plugins {
     id("com.android.library")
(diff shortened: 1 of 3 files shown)


bridge-client/build.gradle.kts
```diff
@@ -27,33 +28,24 @@ kotlin {
...
-        if (iOSTargetName.startsWith("iphoneos"))
-            ::iosArm64
-        else
-            ::iosX64
-    iOSTarget("ios") {
-        binaries {
-            framework {
-                baseName = iosFrameworkName
+    else {
+        iosX64("ios") {
```
Erin-Mounts, line 44: Should you also build for the simulator/arm64 architecture for M1 macs? Or is that not supported yet

Erin-Mounts (approved):
One question, otherwise looks... good? ¯\\_(ツ)\_/¯

bridge-client/build.gradle.kts
syoung-smallwisdom: No, there is a bug where running "build" which is what they use for Android will fail if either (a) both iosArm64 and iosX64 are included *or* if you don't build for either. Has to be simulator only. It's kind of annoying since for the *actual* iOS build, it has to build both sim and device. *sigh*

WRT including macOS (so you can run swift unit tests), that also doesn't work b/c the DateTime library is not supported on macOS. Basically, I am planning on crossing the macOS/tvOS/watchOS bridge later.

PR #65, opened 2021-09-15
Fix loading study

syoung-smallwisdom:
### BMC-219: Use the appState to show launching while loading the study
For whatever reason, iOS 14.4 is not recognizing the changes to the published Study
object. This is the only case where a Kotlin model object is used referenced directly
within SwiftUI. This changes the updateAppState() method to use that to set the
`appState` to `.launching` if the Study is being loaded.

### BMC-218: Logout hangs on loading study 
B/c of how login depends upon the studyId *before* I have all the information about the study
(which is not available to an unauthenticated user), there are a bunch of things that need to be
nil'd out when logging out. Missed some of the ivars.


Erin-Mounts (approved):
Looks good

PR #66, opened 2021-09-15
Update iOS framework build

Erin-Mounts:
previous commit mistakenly built against an older version of the main branch


syoung-smallwisdom (approved):
👍

PR #69, opened 2021-09-17
BMC-224 Fix threading for refresh session token

syoung-smallwisdom:
Kotlin does not allow accessing objects from a different thread from the thread on which
they were created. Wrap all access to the authManager with running it on the main thread.

Fix for https://sagebionetworks.jira.com/browse/BMC-224


Erin-Mounts (approved):
Turns out, since your other changes ensured that retry() is only ever called on the main thread, the "thread-safe" access to the session token isn't needed anyway, so the potential for deadlock goes away.

PR #70, opened 2021-09-17
Show a preview of the PolicyNotice PDF

syoung-smallwisdom:
Showing the share sheet for sharing the PDF is mad slow and
can take *several* seconds to load. This is a work-around
intended to show a preview of the PDF document and gives
the participant the option to print, save to file, air drop, etc.

While this does mean a lot of tapping. Tap to dismiss share then
tap again to dismiss preview, the trade-off is that it loads
**WAY** faster.


Erin-Mounts (approved):
Looks good to me

PR #75, opened 2021-09-24
Fixes to HttpClient headers that break AppConfigApi

liujoshua:


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AppConfigRepo.kt
```diff
@@ -25,7 +27,13 @@ class AppConfigRepo(httpClient: HttpClient,
 
 
     internal var publicApi = PublicApi(
-        httpClient = httpClient
+        httpClient = httpClient.config {
+            defaultRequest {
+                // temporarily hard code Accept-Language to avoid a 404 - liujoshua 2021-09-21
+                // remove after implementing this functionality in HttpClientModule
+                header("Accept-Language", "en-US,en")
```
nategbrown9: Could we do this in HttpClientModule.createBridgeHttpClient() so it gets applied to all requests? Also less likely to miss removing when we put in the generalized solution.

liujoshua: @syoung-smallwisdom will adding `Accept-Language` to the shared HttpClient configuration be okay for iOS? 

NativeAppConfigManager uses the shared API client code. BridgeClientAppManager sets `Accept-Language`. Not sure if these interact at all

nategbrown9 (approved):
Looks good.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/BridgeConfig.kt
```diff
@@ -102,7 +102,7 @@ interface BridgeConfig : PlatformConfig {
             }
             if (info.appName != null) {
                 sb.append("/")
-                sb.append(info.appName)
+                sb.append(info.appVersion)
```
Erin-Mounts: Should the code also be checking that appVersion is not null in the if statement above, or is this intentional?

liujoshua: Good catch. That's what the BridgeJavaClient does but that might cause problems

Looks like there server code handles an app name with no version (at least for the cases I thought of). On the client side, if we don't have device or OS info, we could end up sending `Mobile Toolbox BridgeClientKMM/1` with server parsing the appName as "Mobile Toolbox BridgeClientKMM". 

Looks like there's been an update to [expected formats.](https://developer.sagebridge.org/articles/filtering.html#user-agent-header).

I'm looking at and testing the server and Java client code (which is more or less identical to this)

liujoshua: Looks like for our BridgeConfig, none of the params for ClientInfo would ever be null. 

I'll still check with the server team on whether we should make changes to the Java client or server

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
```diff
@@ -57,9 +57,16 @@ private fun createBridgeHttpClient(enableNetworkLogs: Boolean, bridgeConfig: Bri
         }
     }
 
+    defaultRequest {
+        // temporarily hard code to avoid a 404 when retrieving AppConfig - liujoshua 2021-09-21
```
syoung-smallwisdom, line 61: Nit: Can you mark this comment with a TODO to help track it? Thanks!

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/BridgeConfig.kt
syoung-smallwisdom: Looks like you need to change the whole thing:

```
        if (isNotBlank(info.appVersion) || info.appVersion != null) {
            if (isNotBlank(info.appVersion)) {
                sb.append(info.appVersion)
            }
            if (info.appVersion != null) {
                sb.append("/")
                sb.append(info.appVersion)
            }
        }
```

syoung-smallwisdom: The Obj-c code has this:
```
    NSBundle *mainBundle = [NSBundle mainBundle];
    UIDevice *currentDevice = [UIDevice currentDevice];
    
    NSString *appName = [mainBundle appName];
    NSString *appVersion = [mainBundle appVersion];
    NSString *deviceInfo = [currentDevice deviceInfo];
    
    return [NSString stringWithFormat:@"%@/%@ (%@) BridgeSDK/%0.0f", appName, appVersion, deviceInfo, BridgeSDKVersionNumber];
```

So... I think the it should be this:

```
        val deviceInfo: String = "${info.deviceName}; ${info.osName}/${info.osVersion}"
        return "${info.appName}/${info.appVersion} ($deviceInfo) ${info.sdkName}/${info.sdkVersion}"
```

B/c in Kotlin, a null won't crash and die, it will just print "null", same as Obj-c. Or if you want to get fancy, you could write extensions for the `ClientInfo` to return a default value if `isNullOrEmpty` is true, but that kinda seems overkill. ¯\_(ツ)_/¯

PR #80, opened 2021-10-04
MTB-277 Fix potential race condition

syoung-smallwisdom:
So... I can't repo the issue, but there *may* be a case where something isn't called on the main thread which means the order could result in redisplay? maybe?


Erin-Mounts (approved):
🤷‍♀️ looks good

PR #83, opened 2021-10-07
MTB-336 Set "declined" to true if abandoned

syoung-smallwisdom:
Apparently, the meaning of "declined" should map to "didAbandon" rather than "early exit" (which means they want to finish later, I guess...) and the task end date should only be set if completed *and* not abandoned. *sigh*


Erin-Mounts (approved):
Looks good to me

nategbrown9 (approved):
Looks good. I think that should match the behavior on Android.

PR #86, opened 2021-10-14
Change StudyInfoView to use binding for selected tab

syoung-smallwisdom:
This change is to allow the "contact support" tab (which is not the default tab) to be deep linked.


Erin-Mounts (approved):
looks good to me

PR #88, opened 2021-10-19
Tweek the tab item font size to shrink to fit smaller phones

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/MTB-347


SwiftPackage/Sources/BridgeClientUI/Views/Components/CustomTabView.swift
```diff
@@ -135,6 +135,9 @@ public struct CustomTabView<Tab : TabItem, Content : View>: View {
     // The size of the tab buttons
     private let buttonSize: CGFloat = 64
     
+    // Uses to allow for resizable text.
```
Erin-Mounts: I'm not parsing this. Is there a word missing between "Uses" and "to"?

Erin-Mounts (approved):
One question about a comment, otherwise looks good

PR #90, opened 2021-10-22
Add `isTimeSensitive` flag for indicating that a schedule should interupt "do not disturb"

syoung-smallwisdom:
…

This is about all we can do to support iOS 15 notifications w/o changing the server-side model.

Note: BridgeClient is currently failing to build (and has been for a while) b/c the sign-in unit test is failing. That unit test failure is unrelated to this PR.


Erin-Mounts (approved):
Looks good.

I saw that sign-in unit test failing when I was porting the upload stuff and Nate said it was because Github couldn't get the credentials if your PR was from a different fork. That being said, I later did a PR from a branch of the main Sage fork and saw the same failure so ¯\\\_(ツ)\_/¯

nategbrown9 (approved):
Looks good. We should probably document somewhere the logic for when a notification will interrupt "do not disturb". Looks like the only time we wouldn't, would be for notifications on multi-day sessions, before the final day.

syoung-smallwisdom:
@nategbrown9 
Yeah, this really should be part of the scheduling model. Oh well. I went with "if it's set to expire > 1 day, then it is time sensitive".

PR #92, opened 2021-10-27
Fix BMC-241: Migrate startEventId from SessionInfo to ScheduledSession.

nategbrown9:


syoung-smallwisdom (approved):
Looks good. 

However, before merging, could you please fix the unit test that is failing? Makes it difficult to have confidence that this change still passes the unit tests. I'm not confident about what the intension of the test is so I'm reluctant to fix it. Thanks!

I have attached the log of the failure that I get when running on my machine. It fails at this line:

```
            val signInResult = authRepo.signInExternalId(externalId, externalId)
            assertTrue(signInResult is ResourceResult.Success<*>)
```

[FailureLog.txt](https://github.com/Sage-Bionetworks/BridgeClientKMM/files/7427748/FailureLog.txt)

syoung-smallwisdom (changes requested):
Oh yes, and also please update the xcframework binary by running:

```
./xcframework_build_release.sh
```

This should be run when making any changes to the cross-platform code. Thanks!

syoung-smallwisdom (approved):

PR #94, opened 2021-11-09
Update dependencies and build files so gradle sync completes without errors

nategbrown9:


syoung-smallwisdom (approved):
Looks good...

So... does this mean that SQLDelight now supports building both for device and simulator without failing? That's why the weird code was in there, so that the Android version would build correctly.

Also, does this need an updated xcframework binary? Did you check that `./xcframework_build_release.sh` works?

nategbrown9:
> Looks good...
> 
> So... does this mean that SQLDelight now supports building both for device and simulator without failing? That's why the weird code was in there, so that the Android version would build correctly.
> 
> Also, does this need an updated xcframework binary? Did you check that `./xcframework_build_release.sh` works?

Yes, SQLDelight was updated.
I successfully ran './xcframework_build_release.sh', but didn't check in an updated binary.

syoung-smallwisdom:
> > Looks good...
> > So... does this mean that SQLDelight now supports building both for device and simulator without failing? That's why the weird code was in there, so that the Android version would build correctly.
> > Also, does this need an updated xcframework binary? Did you check that `./xcframework_build_release.sh` works?
> 
> Yes, SQLDelight was updated. I successfully ran './xcframework_build_release.sh', but didn't check in an updated binary.

Thanks!

PR #98, opened 2021-12-03
Add phone sign in apis

nategbrown9:


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
```diff
@@ -58,6 +58,65 @@ class AuthenticationRepository(httpClient: HttpClient, val bridgeConfig: BridgeC
...
     }
 
+    /**
+     *
+     * Will send an SMS message with a code that can be used to call the server and generate a session.
+     * @param number Phone number of participant.
+     * @param regionCode CLDR two-letter region code describing the region in which the phone number was issued.
+     * @return Boolean
+     */
+    suspend fun requestPhoneSignIn(number: String, regionCode: String) : Boolean {
```
syoung-smallwisdom, line 78: What is the format of the `number`? Since Canada and US have the same country code of +1, how does the app determine if the number is US or CA?

syoung-smallwisdom (approved):
Looks good except it would be nice to have some documentation around the formatting of the `number` since it's just a string.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
nategbrown9: I will have to follow up with Alx to make sure it is documented correctly. The documentation from the Rest services is not clear: "The phone number (can be formatted in any way that’s useful for end users)."

liujoshua: Looks like parser is using is https://github.com/google/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java

nategbrown9: The server uses the regionCode to determine how to parse the number. For US or Canada it should be able to handle a 10 digit number with or without the +1, and whatever formatting the user happens to enter.

PR #100, opened 2021-12-07
Fix iOS crash by rethrowing errors when fetching the authorization session

syoung-smallwisdom:
Because the kotlin framework isn't symbolicated, I am not sure why fetching the current session in a background launch is crashing but this will keep that from happening. Note: this might not cause other issues of the upload not occurring in the background but will keep it from crashing. ¯\_(ツ)_/¯ 

https://sagebionetworks.jira.com/browse/MTB-388


bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeAuthenticationManager.kt
```diff
@@ -47,7 +47,12 @@ open class NativeAuthenticationManager(
     }
 
     open fun session() : UserSessionInfo? {
-        return authManager.session()
```
Erin-Mounts, line 50: Yeah, I feel like not being able to get the session token when the app is launched in the background to respond to a background upload event could either be no big deal that it will eventually recover from, or could break it in bad ways resulting in lost data. I've dug through the Swift background upload code and it *looks* like it shouldn't be the latter case, but I'd rather not find out I was wrong about that in production on our first actual studies run by outside partners. @nategbrown9 any ideas why this call would fail/throw an error in that situation? Is there maybe an order of operations we could try to enforce to make sure that doesn't happen?

Erin-Mounts (approved):
Approving because it *should* be ok, with the caveat that if @nategbrown9 has ideas on how to prevent the error case, we should include those for a suspenders-and-belt approach.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeAuthenticationManager.kt
nategbrown9: Do all attempts to access `authManager` fail when launched in the background? Does `BridgeClientAppManager.appWillFinishLaunching` get called when the app is launched in the background? This initializes Koin (dependency injection), which is needed to inject the AuthenticationRepository into NativeAuthenticationManager.

syoung-smallwisdom: So.. the "in the background" is an assumption based on two things: 
1. No reported crashes (from QA or beta testers)
2. Stephanie noticing that she was getting crashes reported over the weekend when she wasn't using the app

Basically, the crash reports are all the same and they all crash during launch when calling for the cached session. Initially, I was rethrowing the error but b/c there is a mock (for testing) that overrides that method, it can't handle a @objc throw with a nil return value (because the participant is not signed in). So I had to swallow the error in the Kotlin code. (Swift doesn't handle exceptions except to crash - you are suppose to throw errors instead.) 

I suppose we could just leave the crash in but I think that this means that the OS won't attempt a re-launch so your upload is going to fail anyway.

WRT whether or not all attempts fail, not sure. Getting the launch in the background to happen is tricky and b/c the Kotlin code doesn't symbolicate, I have no idea what the exception is.

Erin-Mounts: > Does `BridgeClientAppManager.appWillFinishLaunching` get called when the app is launched in the background? This initializes Koin (dependency injection), which is needed to inject the AuthenticationRepository into NativeAuthenticationManager.

As far as I can tell, the *only* time the failing method in question would get called is during this launch/initialization, so I have to assume yes.

ETA I think I'm wrong about that, it also gets called in the observer any time the UserSessionInfo updates. Is there any way it could be getting that observer callback when the NativeAuthenticationManager is in a bad/uninitialized state?

Erin-Mounts: > I suppose we could just leave the crash in but I think that this means that the OS won't attempt a re-launch so your upload is going to fail anyway.

The upload might fail initially, but at least this way would definitely get retried. As far as I can tell that's also the case if you swallow the error in Kotlin, which is why I approved the change anyway, but wanted Nate to weigh in on whether there might be a way to prevent the error. Which there might be if we had a way to see what the error is.... ugh. Anyway, yes, I'm 99% sure it's better to not crash even if it's only happening when they're not using the app.

syoung-smallwisdom: Um... I'm confused, I think.

> ETA I think I'm wrong about that, it also gets called in the observer any time the UserSessionInfo updates. Is there any way it could be getting that observer callback when the NativeAuthenticationManager is in a bad/uninitialized state?

The observer callback does *not* call that method directly. It *sets* the `session` property on the `BridgeClientAppManager` instead. It's also worth noting that the actual background code for getting the session token and uploading isn't crashing. Only when calling `NativeAuthenticationManager.session()` on launch. Since the observer is hooked up directly after this call, it was my reasoning that the session would be set as soon as that observer returned.

```
        // Hook up user session info
        self.authManager = NativeAuthenticationManager() { userSessionInfo in
            guard userSessionInfo == nil || !userSessionInfo!.isEqual(userSessionInfo) else { return }
            self.session = userSessionInfo
        }
        self.session = self.authManager.session()
        self.authManager.observeUserSessionInfo()
```

Screen Shot 2021-12-08 at 9 52 18 AM

syoung-smallwisdom: I suppose we could also just *not* call the `session()` method and always wait until the observer returns?

Erin-Mounts: Hm, that seems like it would be safer (from crashing). I think the background upload code assumes that has been set by the time it tries to upload/retry-upload/tell bridge it finished uploading anything but since it has to switch to the main thread to get the session token, hopefully the observer would have been called by then?

PR #102, opened 2021-12-09
Update timeline model objects to include new StudyBurstInfo.

nategbrown9:


syoung-smallwisdom (approved):
Looks good as a work-in-progress. Because the binaries aren't updated, none of these changes will be available on iOS. That said, it doesn't look like they are currently used?

nategbrown9:
> Looks good as a work-in-progress. Because the binaries aren't updated, none of these changes will be available on iOS. That said, it doesn't look like they are currently used?

Not currently being used. I just was updating the api model objects so they are there when when we need them.

PR #105, opened 2022-01-12
Change to showing the PDF in a webview and then using share

syoung-smallwisdom:
Because the webview will display and *then* load the url, it blocks buttons while loading.
Then there is a button to dismiss and a button to share that shows the slow-to-load share options.

https://sagebionetworks.jira.com/browse/MTB-370

After 7 kajillion different attempts using googled solutions, decided to try a WebView. ¯\\_(ツ)_/¯


Erin-Mounts (approved):
Looks good

PR #106, opened 2022-03-04
Add participant apis and participant repository

nategbrown9:
This adds study participant and data apis along with the new participant schedule API. There is a new ParticipantRepo for clients to update StudyParticipant data. The ParticipantRepo depends on the AuthenticationRepository, but needs to be a separate class to avoid a dependency loop.


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ParticipantRepo.kt
```diff
@@ -0,0 +1,162 @@
...
+    data class UpdateParticipantRecord private constructor (
+        /* An ID assigned to this account by Bridge system. */
+        val id: String,
+        /* First name (given name) of the user. */
+        var firstName: String? = null,
+        /* Last name (family name) of the user. */
+        var lastName: String? = null,
+        var sharingScope: SharingScope? = null,
+        /* The data groups set for this user. Data groups must be strings that are defined in the list of all valid data groups for the app, as part of the app object.   */
+        val dataGroups: MutableList<String>? = null,
```
syoung-smallwisdom, line 116: I think this is a Set? Is there a type for that in Kotlin?

syoung-smallwisdom (approved):
looks good

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ParticipantRepo.kt
nategbrown9: There is, but JSON doesn't support sets. This is just matching the Bridge webservice.

PR #107, opened 2022-03-11
Use append_or_replace work policy and always retry.

nategbrown9:
Google has updated the documentation on append: "When using APPEND with failed or cancelled prerequisites, newly enqueued work will also be marked as failed or cancelled respectively." Using recently added APPEND_OR_REPLACE will start a new chain.

### DIFF
--- a/bridge-client/src/androidMain/kotlin/org/sagebionetworks/bridge/kmm/shared/upload/UploadRequester.kt
+++ b/bridge-client/src/androidMain/kotlin/org/sagebionetworks/bridge/kmm/shared/upload/UploadRequester.kt
@@ -57,6 +57,13 @@ class UploadRequester(
         )
         database.insertUpdateResource(resource)
 
+        queueUploadWorker()
+    }
+
+    /**
+     * Enqueue a WorkManager work request to process uploads.
+     */
+    fun queueUploadWorker() {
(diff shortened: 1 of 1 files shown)


bridge-client/src/androidMain/kotlin/org/sagebionetworks/bridge/kmm/shared/upload/UploadRequester.kt
```diff
@@ -106,7 +113,7 @@ internal class CoroutineUploadWorker(
                 }
             } catch (t: Throwable) {
                 Log.e(TAG, "Error calling uploadManager.ProcessUploads", t)
-                Result.failure()
+                Result.retry()
```
Erin-Mounts, line 116: Will this lead to previous failed uploads being retried even without needing to request a new upload? If so, 👍

Erin-Mounts (approved):
Looks good.

bridge-client/src/androidMain/kotlin/org/sagebionetworks/bridge/kmm/shared/upload/UploadRequester.kt
nategbrown9: Yes, a work request will keep being retried till it gets a success.

PR #109, opened 2022-04-05
Support mac silicon

syoung-smallwisdom:
So... I *think* this allows for using Mac Silicon simulator. Untested on that device since I don't have one. I also fixed the iOS unit tests and binary to match the changes that went into build v0.3.4 and tagged for iOS as 0.3.4


Erin-Mounts (approved):
Looks good to me

nategbrown9 (approved):
Looks good.

PR #110, opened 2022-04-08
Bugfix/bmc 337 notifications

syoung-smallwisdom:
Only return notifications if the session is not complete.


Erin-Mounts (approved):
Looks good to me

nategbrown9 (approved):
Looks good.

PR #111, opened 2022-04-23
Fetch adherence records if this is a new login

syoung-smallwisdom:
Unfortunately, my attempts to fetch the adherence records by triggering after authentication all crashed spectacularly so I hacked something in to the today view model so it will work on iOS. I think this is worth revisiting since the today view model is pretty specific to Lynn's designs of "today" and "next available" where I don't think that works more generally.


nategbrown9:
> 

The crashes were when trying to load adherence records from Swift as part of the sign-in flow?

syoung-smallwisdom:
> 

No, I tried doing something similar in the Kotlin code from within the shared auth manager using the background scope but it didn't like it. In the interest of fixing this *quickly* so I can include it in the other fixes that they want, I figured I'd do it this way and then we could revisit for ARC.

nategbrown9 (approved):
Looks good.

PR #112, opened 2022-04-25
Fix MTB-519: Update user-agent strings to be consistent for both Android and iOS

nategbrown9:


syoung-smallwisdom (approved):
looks good.

Can you check the unit tests before merging?

nategbrown9:
> 

Test failures appear to be transient ones we've seen before.

PR #113, opened 2022-04-25
Gradle and library updates to work with Android Mobile Toolbox App

nategbrown9:
We can't target Android version 31 until Northwestern updates their measures.
Updating WorkManager library so it works with updates in MTB app


syoung-smallwisdom (approved):
Looks good

PR #115, opened 2022-05-10
BridgeFileUploadManager keeps isUploading status flag up-to-date

Erin-Mounts:
…so that it can be used for accurate participant  feedback regarding the progress of result and participant file uploads.

### DIFF
--- a/SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
+++ b/SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
@@ -539,12 +539,10 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
         // Set up a listener to retry temporarily-failed uploads whenever the app becomes active
         NotificationCenter.default.addObserver(forName: UIApplication.didBecomeActiveNotification, object: nil, queue: nil) { notification in
             self.checkAndRetryOrphanedUploads()
-            self.retryUploadsAfterDelay()
         }
         
         // Also do it right away
         self.checkAndRetryOrphanedUploads()
-        self.retryUploadsAfterDelay()
     }
     
     fileprivate func touch(fileUrl: URL) {
@@ -618,8 +616,17 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
         return tempFileURL
     }
     
+    fileprivate func runOnQueue(_ queue: OperationQueue, sync: Bool, block: @escaping ()->Void) {
+        if OperationQueue.current == queue {
+            block()
+        } else {
+            queue.addOperations([BlockOperation(block: block)], waitUntilFinished: sync)
+        }
+    }
+    
+    
     fileprivate func persistMapping<T>(from key: String, to value: T, defaultsKey: String) where T: Encodable {
-        let block = {
+        self.runOnQueue(self.uploadQueue, sync: false) {
             var mappings = self.userDefaults.dictionary(forKey: defaultsKey) ?? [String : Any]()
             var plistValue: Any
             do {
@@ -631,19 +638,12 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
             mappings[key] = plistValue
             self.userDefaults.setValue(mappings, forKey: defaultsKey)
         }
-        
-        if OperationQueue.current == self.uploadQueue {
-            block()
-        }
-        else {
-            self.uploadQueue.addOperation(block)
-        }
     }
     
     @discardableResult
     func removeMapping<T>(_ type: T.Type, from key: String, defaultsKey: String) -> T? where T: Decodable {
         var mapping: T?
-        let block = {
+        self.runOnQueue(self.uploadQueue, sync: true) {
             var mappings = self.userDefaults.dictionary(forKey: defaultsKey)
             if let mappingPlist = mappings?.removeValue(forKey: key) {
                 do {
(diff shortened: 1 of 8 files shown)


SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -1001,14 +975,24 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
         return notCanceledTasks
     }
     
-    /// Call this function to check for and retry any orphaned uploads.
-    public func checkAndRetryOrphanedUploads() {
+    fileprivate func fileIsADayOld(invariantFilePath: String, fileUrl: inout URL!) -> Bool {
+        let oneDay: TimeInterval = 60 * 60 * 24
+        let fullPath = self.fullyQualifiedPath(of: invariantFilePath)
+        fileUrl = URL(fileURLWithPath: fullPath)
+        guard let modDate = self.modificationDate(of: fileUrl) else { return false }
```
syoung-smallwisdom, line 982: So this checks the mod date on the file, right? What if the file fails to upload for 2 days b/c the person is offline? Then will the checkForOrphanedUploads get into an infinite loop of wacky madness? Or is this just a "don't worry about it for the first day" and if so... why?

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -1140,6 +1120,50 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
         }
     }
     
+    /// Call this function to check for and retry any orphaned uploads, and update the app's isUploading state  accordingly.
+    public func checkAndRetryOrphanedUploads() {
+        // This needs to start out from the main queue to avoid an EXC_BAD_ACCESS crash in
+        // checkForOrphanedUploads(), but we also need that function to do its thing before
+        // continuing on here and we don't want to potentially deadlock by popping out
+        // synchronously to main in there. So we'll do it from here.
```
syoung-smallwisdom, line 1128: Um... huh?

SwiftPackage/Tests/BridgeClientUITests/BridgeFileUploadManagerTestCase.swift
```diff
@@ -247,81 +247,82 @@ extension BridgeFileUploadManagerTestCaseTyped {
...
-                }
-                
                 // -- do the retry
                 bfum.retryUploadsAfterDelay()
             }
         }
         
         // -- wait here until everything above has worked its way through.
-        XCTWaiter(delegate: self).wait(for: [expectRetried], timeout: 5.0)
+        XCTWaiter(delegate: self).wait(for: [expectRetried], timeout: 500.0)
```
syoung-smallwisdom: So... if this really takes 8 minutes (b/c you're queuing up "a bunch of times") then we will want to revisit how often this test is run so that we can use the unit tests to test other code paths. Can you add a TODO for that?

SwiftPackage/Tests/BridgeClientUITests/Resources/pf-upload-request-success.json
```diff
@@ -2,5 +2,6 @@
     "fileId":"TestFileId",
     "mimeType": "image/jpeg",
     "uploadUrl":"/not-a-real-pre-signed-S3-url",
+    "expiresOn":"2999-12-31T11:59:59.999Z",
```
syoung-smallwisdom, line 5: In the year 2999, I don't know if man will still be alive...

syoung-smallwisdom (approved):
Couple of questions about things I don't entirely grok. Otherwise, looks good.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: This basically is to have it wait at least one day before retrying an orphaned file upload to minimize the (unavoidable, unfortunately, but eventually recoverable) possibility of stomping on the state of a still-in-progress attempt that only *looks* orphaned because it happens to be doing this check in between two steps of updating state flags that should ideally be done together atomically but can't for complicated reasons involving needing to access the session token via Kotlin from the main queue while handling requests/responses on the URLSession's background queue.

It won't get into an infinite loop of wacky madness if they're offline for 2 days because (barring that weird timing issue, which this will make extremely rare) it won't even consider the file to be orphaned if the corresponding background upload/download task is still in the URLSession's list of tasks waiting for an opportunity to connect and complete.

SwiftPackage/Sources/BridgeClientUI/File Uploading/BridgeFileUploadManager.swift
Erin-Mounts: checkForOrphanedUploads() needs to start from the main queue to avoid an EXC_BAD_ACCESS crash. This method needs the results before it can proceed. If I have checkForOrphanedUploads() always put itself onto the main queue, synchronously, to avoid the EXC_BAD_ACCESS no matter where it's called from, then if this method were called from the main queue (which it sometimes is) it would deadlock. But the caller of this method doesn't need it to complete before proceeding, so instead I have this method toss itself *asynchronously* onto main before proceeding, and have checkForOrphanedUploads()'s documentation comments say that it needs to be called on the main queue only.

Concurrency and asynchronosity(?) are hella confusing, which is why it took me so #$%&@ long to sort all this out.

SwiftPackage/Tests/BridgeClientUITests/BridgeFileUploadManagerTestCase.swift
Erin-Mounts: Oops. I had bumped it up so I could have some time to poke around debugging stuff before it timed out on me. I've bumped it back down to 5 seconds which is more than plenty when not stopping at breakpoints.

SwiftPackage/Tests/BridgeClientUITests/Resources/pf-upload-request-success.json
Erin-Mounts: thanks for *that* earworm.

PR #122, opened 2022-06-09
Assorted fixes to Swift package discovered while building BiAffect preview app

syoung-smallwisdom:


Erin-Mounts (approved):
Looks good to me

PR #125, opened 2022-06-24
Switch base64 encoder libraries for md5 hash

nategbrown9:
Previous library would sometimes put in a newline character, causing https://sagebionetworks.jira.com/browse/BMC-395. The okio documentation explicitly says it does not use newline characters.


syoung-smallwisdom (approved):
looks good

PR #126, opened 2022-07-08
Workaround to allow ParticipantSchedule to be modified before being cached locally.

nategbrown9:


dephillipsmichael:
Thank you! Looks like exactly what I need.

syoung-smallwisdom (approved):
looks good

PR #127, opened 2022-07-09
Hook needed for iOS

syoung-smallwisdom:
Need to add accessor and also need to build the binary using `./xcframework_build.sh`


nategbrown9 (approved):
Looks good.

PR #130, opened 2022-07-20
Add createActivityEvent api

nategbrown9:
This adds the createActivityEvent api to the kotlin code and allows it to be used on Android. This does not include support for iOS yet.

### DIFF
--- a/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/ActivityEventsApi.kt
+++ b/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/ActivityEventsApi.kt
@@ -5,7 +5,9 @@ package org.sagebionetworks.bridge.kmm.shared.apis
 
 
 import io.ktor.client.*
+import org.sagebionetworks.bridge.kmm.shared.models.Message
 import org.sagebionetworks.bridge.kmm.shared.models.StudyActivityEventList
+import org.sagebionetworks.bridge.kmm.shared.models.StudyActivityEventRequest
 
 internal class ActivityEventsApi(basePath: String = BASE_PATH, httpClient: HttpClient) : AbstractApi(basePath, httpClient) {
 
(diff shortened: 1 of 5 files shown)


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/models/StudyActivityEventRequest.kt
```diff
@@ -0,0 +1,27 @@
...
+    /* ISO 8601 date and time that the event occurred */
+    @SerialName("timestamp")
+    val timestamp: Instant,
+    /* answer value for a question event */
+    @SerialName("answerValue")
+    val answerValue: String? = null,
+    /* An IANA Time Zone Database (TZDB) name for the participant’s time zone at the time the record is submitted.
+     If present, this field will be validated to be a valid time zone name. */
+    @SerialName("clientTimeZone")
+    val clientTimeZone: String? = null,
```
dephillipsmichael, line 22: I keep running into the "IANA Time Zone Database (TZDB)" for Bridge calls, how do you actually get the IANA Timezone name of a user on Android?

dephillipsmichael:
@nategbrown9 looks good, should unblock me.  thanks for the quick turn-around.

dephillipsmichael (approved):

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/models/StudyActivityEventRequest.kt
nategbrown9: With kotlinx.datetime you can do: TimeZone.currentSystemDefault().toString()

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/ActivityEventsApi.kt
```diff
@@ -19,4 +21,19 @@ internal class ActivityEventsApi(basePath: String = BASE_PATH, httpClient: HttpC
...
 
+    /**
+     * Create a new activity event for the caller in this study. This call returns 201 whether the event updates or not,
+     * to support concurrent and out-of-order POSTs from the client. To return 400 if an event fails to publish,
+     * set the showError parameter to true. The creation or update of an event that triggers a study burst will update
+     * the study burst events as well, even to the point of recreating study burst events that have been deleted
+     * (depending on the update type of the bursts). To decouple these events and update the origin event without
+     * changing the study burst events, set the updateBursts parameter to false.
+     * Client is using server defaults:
+     * showError = false
```
Erin-Mounts, line 32: Should these be parameters to the function, then, with default values as shown if not passed in explicitly as arguments? Or does Kotlin not support default parameter values?

Erin-Mounts (approved):
One question, otherwise looks good

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/ActivityEventsApi.kt
nategbrown9: This is mostly copied from the server api documentation. At the moment I think the default behavior should be all we need, but I wanted to document it here in case we ever need to change it.

PR #132, opened 2022-07-27
Add ability to set UserSessionInfo to support migrating from old Bridge client libraries.

nategbrown9:


syoung-smallwisdom (approved):
Looks good. Will need to update native and binary to use on iOS.

dephillipsmichael:
Looks good to me too.  I like the check to make sure the session is currently null.

PR #133, opened 2022-07-28
build for app extension

syoung-smallwisdom:
The Kotlin binary was not setting the compile flag for linking as an app extension. I also did a code hunt and wrapped an import of UIKit in case we ever support other platforms than iOS.


Erin-Mounts (approved):
Looks good to me.

nategbrown9 (approved):
Looks good.

PR #134, opened 2022-08-09
Stub out abstract NativeTimelineManager

syoung-smallwisdom:
This is to allow Michael to use shared code in commonMain for getting and caching the sessions as needed to support the showing study bursts. Since there is more work to support that feature, I am not bumping the version or building a binary for this change.

@dephillipsmichael

### DIFF
--- a/bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
+++ b/bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
@@ -19,15 +19,7 @@ class NativeTimelineManager(
     private val includeAllNotifications: Boolean,
     private val alwaysIncludeNextDay: Boolean,
     private val viewUpdate: (NativeScheduledSessionTimelineSlice) -> Unit
-) : KoinComponent {
-
-    private val repo : ScheduleTimelineRepo by inject(mode = LazyThreadSafetyMode.NONE)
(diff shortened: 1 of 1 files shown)


nategbrown9 (approved):
Looks good.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -49,6 +41,19 @@ class NativeTimelineManager(
         runCatching { scope.cancel() }
         observeTodaySchedule(false)
     }
+}
+
+abstract class AbstractNativeTimelineManager(
+    private val studyId: String
+) : KoinComponent {
+
+    internal val repo : ScheduleTimelineRepo by inject(mode = LazyThreadSafetyMode.NONE)
```
dephillipsmichael, line 50: I do need access to the ScheduleTimelineRepo, or the ResourceDatabaseHelper, for a DIAN Android app feature that doesn't belong in the shared library.  Can we make all of these val's public instead of internal?

dephillipsmichael: Since this was merged, I'll just update my PR

syoung-smallwisdom: yeah, but this is the "native" version that isn't used on Android and is only defined in the iOS framework. :)

PR #141, opened 2022-08-18
study_bursts

dephillipsmichael:
@nategbrown9 and @syoung-smallwisdom for review. Also, how do I re-build this for iOS? (Will need done before merging.)

**CommonMain:**
1) Refactored the `getPastSessions` function to `getSessions` to allow for the future session function to share the code.
2) Added the ability for the `ScheduleTimelineRepo` to get future sessions.  Added unit tests for a few scenarios.
3) Added the ability to get a new data model `StudyBurstSchedule` which is a user's entire timeline organized into a sorted list of `StudyBurst` with sessions grouped day.

**iOSMain**
1) Added the `NativeTimelineStudyBurstManager` which allows for loading the `StudyBurstSchedule`
2) Added functionality to the `AbstractNativeTimelineManager` which allows for loading the full list of AdherenceRecords, and re-running the schedule mutator.


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+     * @param timezone provide a custom timzone if desired, i.e. for testing.
+     */
+    fun getStudyBurstSchedule(studyId: String,
+                              userJoinedDate: Instant,
+                              timezone: TimeZone = TimeZone.currentSystemDefault())
+            : Flow<ResourceResult<StudyBurstSchedule>> {
+
+        val oneDayBeforeJoin = userJoinedDate.minus(1, DateTimeUnit.DAY, timezone)
+        return getFutureSessions(studyId, oneDayBeforeJoin).map {
+            when (it) {
```
dephillipsmichael, line 157: Is this the best way to convert from one ResourceResult to another generic type ResourceResult?

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
         }
     }
 
+    /**
+     * Get all sessions with a start time after the date provided for now.
+     * @param studyId the study identifier
+     * @param now the current instant, or what you want to provide in testing.
+     */
+    fun getFutureSessions(studyId: String, now: Instant = Clock.System.now())
```
syoung-smallwisdom: complete nit but since in `getStudyBurstSchedule` you are using this method with a "from" date that is not "now" but still valid, maybe it would be better to rename that parameter?

The "now" parameter is included in this code file so that unit tests can set that date to avoid test failures due to crossing the midnight time/date border. Perfectly reasonable that it might not always be "now" but I'd like to see it renamed so as to avoid confusion later. :)

```suggestion
    fun getFutureSessions(studyId: String, fromInstant: Instant = Clock.System.now())
```

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+        : Flow<ResourceResult<ScheduledSessionTimelineSlice>> {
+
+        return getTimeline(studyId).map {
+            extractFutureSessionsFromResults(it, studyId, now)
```
syoung-smallwisdom:
```suggestion
            extractFutureSessionsFromResults(it, studyId, fromInstant)
```

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+     * @param studyId Study identifier
+     * @param userJoinedDate the instant the user joined the study
+     * @param timezone provide a custom timzone if desired, i.e. for testing.
+     */
+    fun getStudyBurstSchedule(studyId: String,
+                              userJoinedDate: Instant,
+                              timezone: TimeZone = TimeZone.currentSystemDefault())
+            : Flow<ResourceResult<StudyBurstSchedule>> {
+
+        val oneDayBeforeJoin = userJoinedDate.minus(1, DateTimeUnit.DAY, timezone)
```
syoung-smallwisdom, line 155: Can you add a comment for future readers as to why you are getting the study burst from a day before the user joined. Thanks!

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+                    ResourceResult.InProgress
+                }
+            }
+        }
+    }
+
+    private fun extractFutureSessionsFromResults(
+        timelineResult: ResourceResult<ParticipantSchedule>,
+        studyId: String,
+        userJoinedDate: Instant,
```
syoung-smallwisdom:
```suggestion
        instantInDay: Instant,
```

See comment above and match naming convention in method below.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -124,16 +126,74 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
...
+    ): ResourceResult<ScheduledSessionTimelineSlice> {
+        return extractSessionsFromResults(timelineResult, studyId, userJoinedDate, WindowFilterType.Future)
```
syoung-smallwisdom:
```suggestion
        return extractSessionsFromResults(timelineResult, studyId, instantInDay, WindowFilterType.Future)
```

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -164,7 +225,10 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
         val schedules = mutableListOf<ScheduledSessionWindow>()
 
         windows.forEach { window ->
-             if (window.startDateTime <= today) {
+             if (filterType == WindowFilterType.Future &&
+                 window.startDateTime > today) {
+                 schedules.add(window)
+             } else if (window.startDateTime <= today) {
                  schedules.add(window)
              }
```
syoung-smallwisdom:
```
if (window.startDateTime <= today || filterType == WindowFilterType.Future) {
    schedules.add(window)
}
```

Doesn't this do the same thing?

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/StudyBurstScheduleJsonResource.kt
```diff
@@ -0,0 +1,14 @@
+package edu.wustl.dian.app.schedule
```
syoung-smallwisdom: can you please change the package name? Thanks!

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -64,6 +91,16 @@ abstract class AbstractNativeTimelineManager(
         }
     }
 
+    fun getAllCachedAdherence(callBack: (List<NativeAdherenceRecord>) -> Unit) {
```
syoung-smallwisdom: We avoid "get" in the iOS code because that word already has a different meaning. Below when getting the AssessmentConfig we use "fetch" to indicate that there is a callback.

That said, I'm not sure why you are adding this here. Not certain what your UI/UX is that requires it, but this is the purpose of the `NativeHistoryManager`.

```suggestion
    fun fetchAllCachedAdherence(callBack: (List<NativeAdherenceRecord>) -> Unit) {
```

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -14,6 +16,31 @@ import org.sagebionetworks.bridge.kmm.shared.models.*
 import org.sagebionetworks.bridge.kmm.shared.repo.*
 import platform.Foundation.*
 
+class NativeTimelineStudyBurstManager(
+    private val studyId: String
+) : AbstractNativeTimelineManager(studyId) {
+
+    fun getStudyBurstSchedule(isNewLogin: Boolean,
```
syoung-smallwisdom: See comment below about "get" vs. "fetch" - you have a callback so should be fetch.

That said, why isn't this using an observer allow refreshing your view with updated adherence records or changes to the event dates?

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -14,6 +16,31 @@ import org.sagebionetworks.bridge.kmm.shared.models.*
...
+                              userJoinedDate: Instant,
+                              scheduleMutator: ParticipantScheduleMutator?,
+                              completed: (NativeStudyBurstSchedule?) -> Unit) {
+        scope.launch {
+            if (isNewLogin) {
+                adherenceRecordRepo.loadRemoteAdherenceRecords(studyId)
+            }
+            repo.scheduleMutator = scheduleMutator
```
syoung-smallwisdom: I feel like maybe we are shooting ourselves in the foot with this setter, especially on a "fetch" rather than an "observe". Maybe this belongs in the constructor?

syoung-smallwisdom (commented):
Lots of questions and nits.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
syoung-smallwisdom: @nategbrown9 ?

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
dephillipsmichael: I did switch it over to an observer.  I tried to add it to the constructor, but I'm not sure it will work since the call repo.scheduleMutator has repo injected.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -14,6 +16,45 @@ import org.sagebionetworks.bridge.kmm.shared.models.*
...
 
+class NativeTimelineStudyBurstManager(
+    private val studyId: String,
+    private val viewUpdated: (NativeStudyBurstSchedule?, String?) -> Unit,
+    private val scheduleMutator: ParticipantScheduleMutator?,
+    private val updateFailed: (() -> Unit)? = null,
+) : AbstractNativeTimelineManager(studyId) {
+
+    init {
+        repo.scheduleMutator = scheduleMutator
```
dephillipsmichael: @syoung-smallwisdom Do you think this will work with the lazy injection?

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -164,9 +227,9 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
         val schedules = mutableListOf<ScheduledSessionWindow>()
 
         windows.forEach { window ->
-             if (window.startDateTime <= today) {
-                 schedules.add(window)
-             }
+            if (window.startDateTime <= today || filterType == WindowFilterType.Future) {
```
dephillipsmichael, line 230: @syoung-smallwisdom I re-ran the unit tests with this logic and it worked as well.  Good suggestion.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
dephillipsmichael: Switched over to the observe pattern.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
dephillipsmichael: Switched to observer

SwiftPackage/Binaries/debug/BridgeClient.xcframework/ios-arm64_x86_64-simulator/BridgeClient.framework/Headers/BridgeClient.h
```diff
@@ -2588,85 +2594,6 @@ __attribute__((swift_name("ResourceType")))
...
-- (NSString *)description __attribute__((swift_name("description()")));
-@property (readonly) NSString *endDate __attribute__((swift_name("endDate")));
-@property (readonly) NSString *instanceGuid __attribute__((swift_name("instanceGuid")));
-@property (readonly) NSString * _Nullable sessionJson __attribute__((swift_name("sessionJson")));
-@property (readonly) NSString *startDate __attribute__((swift_name("startDate")));
-@property (readonly) NSString *startTime __attribute__((swift_name("startTime")));
-@property (readonly) NSString *studyId __attribute__((swift_name("studyId")));
-@end;
-
-__attribute__((objc_subclassing_restricted))
```
dephillipsmichael, line 2640: @syoung-smallwisdom it seems odd to me that these were removed. I don't remember touching these files.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
syoung-smallwisdom: right... forgot that's why we did that.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
syoung-smallwisdom: I'm not sure. You might want to add to the Swift unit tests to check. Also, if you are going to have the schedule mutator in the init, shouldn't it be on the abstract timeline manager?

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -172,13 +220,35 @@ internal fun ScheduledNotification.toNative()  =
...
     val notifications: List<NativeScheduledNotification>,
 )
 
+data class NativeStudyBurstSchedule (
+    val timezone: NSTimeZone,
+    val studyBurstList: List<NativeStudyBurst>
+)
+
+data class NativeStudyBurst (
+    val sessions: List<List<NativeScheduledSessionWindow>>
```
syoung-smallwisdom: the nested array is a little confusing. Could you flatten it?

syoung-smallwisdom (commented):
Now that you have the binary built, can you please add some unit tests to check that it doesn't crash the init with the mutator? -- OR -- since it's an observer, it also makes sense in that case to put initialize on the observer since that only gets re-added if a new timeline manager is needed.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
dephillipsmichael: I moved it to the abstract timeline manager constructor, and I added the mutator call to your NativeTimelineManagerTests in the iOS App project

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
dephillipsmichael: It looks like the constructor method works and is called after the injection.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
nategbrown9: That's what we did elsewhere in the code. I haven't figured a more elegant solution.

nategbrown9 (approved):
Looks good.

syoung-smallwisdom (approved):

PR #143, opened 2022-08-22
Refactor bridge client packages

syoung-smallwisdom:
I separated out refactoring the Swift package and using an app group so as to unblock @dephillipsmichael


Erin-Mounts (approved):
Looks good to me

PR #144, opened 2022-08-23
Fix metadata map serialization

nategbrown9:


syoung-smallwisdom (approved):
Looks... ok.

Can we plan to revisit this in September to use serialized data classes instead for both the file info objects and the metadata object. Key-value pairs make me nervous.

dephillipsmichael:
Just my two cents, but I agree it would be better as a data class.  You could have the data class contain a Map<String, String> for "extras" that individual apps could use, similar to what you had before.

nategbrown9:
Agree with both comments. This is on my list of things I want to tackle in September.

PR #145, opened 2022-08-23
App Extension

syoung-smallwisdom:
This updates the constructor for the database to put it in the shared group if there is an app group identifier.

It also fixes uploading of the archive for the case where there is a shared group that is used as the temp storage for the upload archive. Turns out that that code was broken and only worked for the case where there is no app group b/c the sim links did not include "private".

Ideally, I'd like to figure out a better way to test uploading than to use existing apps and adhoc testing but this *should* unblock BiAffect (though I have a lot more testing to do of the keyboard extension).


Erin-Mounts (approved):
looks good

PR #149, opened 2022-09-14
Define dependencies using version catalog

nategbrown9:
This uses Gradle's new [version catalog](https://docs.gradle.org/current/userguide/platforms.html) functionality for centrally declaring dependencies in multi-module projects. Functionally this is the same as what we were doing with the buildSrc module, but more efficient and should be better supported going forward.


syoung-smallwisdom (approved):
Looks good? Can you add to the PR description what this does? Guessing its to simplify dependency management to avoid stuff like MTB depending upon a different library version by allowing ranges?

PR #150, opened 2022-09-16
Fix BMC-405: Update AssessmentResultArchiveUploader metadata map to use a data class

nategbrown9:


assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
```diff
@@ -208,13 +197,31 @@ abstract class AssessmentResultArchiveUploader(
...
+    val startDate: kotlinx.datetime.Instant,
+    val endDate: kotlinx.datetime.Instant?,
+    val files: List<ArchiveFileInfo>
+)
+
+@Serializable
+data class ArchiveFileInfo(
+    val filename: String,
+    val timestamp: String,
+    val contentType: String? = null
```
syoung-smallwisdom:
```suggestion
    val contentType: String? = null
    val identifier: String? = null
    val stepPath: String? = null
    val jsonSchema: String? = null
```

This would match the Swift model:
```swift
public struct FileInfo : Codable, Hashable, Equatable {

    /// The filename of the archive object. This should be unique within the manifest. It may include
    /// a relative path that points to a subdirectory.
    public let filename: String
    
    /// The file creation date.
    public let timestamp: Date
    
    /// The content type of the file.
    public let contentType: String?
    
    /// The identifier for the result. This value may *not* be unique if a step is run more than once
    /// during a task at different stages.
    public let identifier: String?
    
    /// The full path to the result if it is within the step history.
    public let stepPath: String?
    
    /// The uri for the json schema if the content type is "application/json".
    public let jsonSchema: URL?
}
```

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
```diff
@@ -208,13 +197,31 @@ abstract class AssessmentResultArchiveUploader(
...
+    val deviceInfo: String,
+    val appName: String,
+    val appVersion: String,
+    val deviceTypeIdentifier: String,
+    val taskRunUUID: String,
+    val dataGroups: String,
+    val startDate: kotlinx.datetime.Instant,
+    val endDate: kotlinx.datetime.Instant?,
+    val files: List<ArchiveFileInfo>
+)
```
syoung-smallwisdom, line 220: Note: A lot of these properties are deprecated. It's not a problem that they are include, but they should be nullable. That said, I don't remember why the other values were removed from the metadata schema. I think it was b/c the archive isn't split up into different archives in order to map to different tables anymore and a lot of this information is in the "taskResult.json" file?

https://github.com/Sage-Bionetworks/mobile-client-json/blob/main/schemas/v2/ArchiveMetadata.json

```
@Serializable
 data class MetadataMap(
     val deviceInfo: String,
     val appName: String,
     val appVersion: String,
     val deviceTypeIdentifier: String,
     val files: List<ArchiveFileInfo>
 )
```

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
```diff
@@ -208,13 +197,31 @@ abstract class AssessmentResultArchiveUploader(
...
 
-fun ArchiveFile.fileInfoMap() : Map<String, JsonElement> {
-    val map = mutableMapOf(
-        "filename" to JsonPrimitive(filename),
-        "timestamp" to JsonPrimitive(endDate.toString())
-    )
+fun ArchiveFile.fileInfo(): ArchiveFileInfo {
     if (this is JsonArchiveFile) {
-        map.put("contentType", JsonPrimitive("application/json"))
+        return ArchiveFileInfo(filename, endDate.toString(), "application/json")
```
syoung-smallwisdom, line 202: Will `toString()` format to the expected serialization format?

syoung-smallwisdom (approved):
Need to add some properties and set the results as nullable.

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
nategbrown9: Yes

PR #152, opened 2022-09-23
Fix using a shared app extension

syoung-smallwisdom:
Uploads to S3 from the app extension still isn't working but this fixes the uploads from the main app, adherence record scoring, and using a shared session token.


Erin-Mounts (approved):
Looks good.

PR #154, opened 2022-09-27
Fix s3 uploads

syoung-smallwisdom:
The method for uploading and encrypting a collection of `DataArchive` objects is older code included to support SageResearch assessments that rely upon retaining the `TaskState` and using that to handle building the archive in the background and then cleaning up and releasing the memory.

The newer assessments built using AssessmentModel use the app manager (which is a singleton) to handle retaining the results and pointers to output directories, etc. to handle archiving the files and then cleaning up the temp data. They use async/await and expect to have the `buildArchive()` method called by the app manager, but b/c they were set up to inherit from `DataArchive`, they conformed to the `encryptAndUpload(_ archives: [DataArchive])` method which does *not* build the archive. I was accidentally calling the wrong method when uploading keyboard data which resulted in empty zip files.

Oops.

I'm not sure how to test for this but the next best thing I could think of was to *not* use that inheritance pattern.

@dephillipsmichael This shouldn't effect your code (even if you updated to a newer version of the library) but figured you should know about it. :)

@nategbrown9  While spelunking, I also found code in the kotlin library that I commented out. It is never called on iOS but if it's used on Android then it would be returning a different userAgent string than the cross-platform one used for most service calls.


Erin-Mounts (approved):
Looks good

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/BridgeConfig.kt
```diff
@@ -55,74 +55,75 @@ interface BridgeConfig : PlatformConfig {
...
+//            }
+//        }
+//        return if ("" == sb.toString()) {
+//            null
+//        } else sb.toString()
+//    }
+//
+//    private fun isNotBlank(string: String?): Boolean {
+//        return string != null && string.isNotEmpty()
+//    }
```
nategbrown9: Just go ahead and delete it. I think it was a copy paste from BridgeAndroidSK that never got cleaned up.

```suggestion

```

PR #158, opened 2022-10-21
DIAN-548

dephillipsmichael:
@syoung-smallwisdom @nategbrown9 see Jira issue https://sagebionetworks.jira.com/browse/DIAN-548

Store and cancel jobs when refreshing callback connections. I'm hoping that cancelling the job specifically will allow iOS to call refreshStudyBurstSchedule and have it work as expected.


bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -86,7 +91,7 @@ class NativeTimelineManager(
     }
 
     fun refreshTodaySchedule() {
-        runCatching { scope.cancel() }
+        runCatching { todayJob?.cancel() }
```
nategbrown9, line 95: Could then set todayJob = null so cancel doesn't get called on it again.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/NativeTimelineManager.kt
```diff
@@ -25,14 +26,16 @@ class NativeTimelineStudyBurstManager(
     private val updateFailed: (() -> Unit)?
 ) : AbstractNativeTimelineManager(studyId, scheduleMutator) {
 
+    var scheduleJob: Job? = null
+
     fun refreshStudyBurstSchedule(userJoinedDate: Instant) {
-        runCatching { scope.cancel() }
+        runCatching { scheduleJob?.cancel() }
```
nategbrown9, line 32: Could then set todayJob = null so cancel doesn't get called on it again.

nategbrown9 (approved):
Couple minor comments, otherwise looks good.

dephillipsmichael:
@nategbrown9 I re-built the iOS library, and I also tested my changes locally, and they work!  I was able to refresh the schedule from bridge without having to force close the app.

syoung-smallwisdom (approved):
looks good

PR #159, opened 2022-10-21
Fix for BMC-419: needSave flag being incorrectly set when loading adherence records.

nategbrown9:


syoung-smallwisdom (approved):
looks good

PR #161, opened 2022-10-24
Fix BMC-420: Re-auth network errors should not clear a user's session.

nategbrown9:


syoung-smallwisdom (approved):
"It'll do" (smile)

What's the code for "not consented"? I thought that was 413?

Honestly, I feel like this is a good case for using an array of status codes - potentially one that could be added to w/o having to update to a new version of the library?

Also, need to also update the iOS binary, yeah?

nategbrown9:
> "It'll do" (smile)
> 
> What's the code for "not consented"? I thought that was 413?
> 
> Honestly, I feel like this is a good case for using an array of status codes - potentially one that could be added to w/o having to update to a new version of the library?
> 
> Also, need to also update the iOS binary, yeah?

Not consented is 412. We don't want to clear a user's session on a 412, because for some app based consent flows you would first need to authenticate before you could call the consent endpoint.

Do need new iOS binaries, but there are several PRs inflight right now that I would like to get merged before updating the frameworks.

PR #165, opened 2022-10-27
Added hook-up to participant reports for iOS

syoung-smallwisdom:
I would say that there is room for improvement on this but... ¯\_(ツ)_/¯


nategbrown9 (approved):
Looks good.

PR #166, opened 2022-10-28
Update Swift Package to support JsonModel “1.6.0”..<“3.0.0”

syoung-smallwisdom:
@dephillipsmichael FYI - this moves around some code. If you are using any of the `JsonModel.ResultData` stuff already then you may want to have a quick tour of it and you may need to add a few import statements so that you can migrate your code to JsonModel v2


Erin-Mounts (approved):
👍

PR #171, opened 2022-11-16
Make the “today” and “history” cards public

syoung-smallwisdom:
By making the today and history cards public, I can preview icons in the app that includes them. Additionally, I can use those features in views that do not include the same "today + next up" filtering, etc.

I also fixed a bug with the default survey icon being the title page instead of the list version of the image and I allow the app to set a default color for the cards.


Erin-Mounts (approved):
looks good

PR #174, opened 2022-12-06
Add imageResource field to AssessmentInfo

nategbrown9:


syoung-smallwisdom (changes requested):
This will break iOS once the binary in built. iOS init doesn't support default value. Can you update that as well? Thanks!

nategbrown9:
> This will break iOS once the binary in built. iOS init doesn't support default value. Can you update that as well? Thanks!

Updated binaries and tests so that swift tests now pass.

syoung-smallwisdom (approved):

PR #176, opened 2022-12-14
Fix check for current login state to work for already logged in participant

syoung-smallwisdom:
… Dangit


Erin-Mounts (approved):
looks good

PR #179, opened 2023-01-06
Add consent as work-around for setting a default study id

syoung-smallwisdom:
This implements the flow of "sign up the participant" and then "consent them to a default study" for email login.

This *should* work for both in-app consent and registration and for consent and registration via email for web-based sign up, but it is currently only implemented for iOS. 

Testing was done with the BiAffect-3 app and using swagger.


Erin-Mounts (approved):
looks good

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ParticipantRepo.kt
```diff
@@ -63,6 +62,27 @@ class ParticipantRepo(httpClient: HttpClient,
...
+            val name = sessionInfo?.let {
+                listOfNotNull(it.firstName, it.lastName).joinToString(" ").ifEmpty { null }
+            } ?: "Name"
+            val scope = sessionInfo?.sharingScope ?: SharingScope.SPONSORS_AND_PARTNERS
+            val consentSignature = ConsentSignature(name = name, scope = scope)
+            val userSession = participantApi.createConsentSignature(subpopulationGuid, consentSignature)
+            authenticationRepo.updateCachedSession(null, userSession)
+            return ResourceResult.Success(userSession, ResourceStatus.SUCCESS)
+        } catch (err: Throwable) {
+            database.removeResource(
```
nategbrown9: I don't think you want to remove the user session resource on a consent failure. This would force the user to sign in again before being able to retry the consent call.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
```diff
@@ -190,16 +190,16 @@ class AuthenticationRepository(
     suspend fun signUpEmail(email: String,
                             password: String,
                             testUser: Boolean = false,
+                            name: String? = null,
```
nategbrown9, line 193: Should this be 'lastName' since that is what it is mapped to? Do we also want to expose 'firstName'?

syoung-smallwisdom: The whole "first name/last name" is a serious pain and this is how we worked around it for years. Even the concept of "first" and "last" name is weird. I don't really want to expose that that's how we are saving the name for signing the consent just b/c we didn't use the same convention for these services. ¯\_(ツ)_/¯

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ParticipantRepo.kt
syoung-smallwisdom: I'll move it to the sign in code - need to ensure that the consent is signed or else the participant doesn't get a schedule b/c they aren't in any studies.

Erin-Mounts (approved):
looks good

nategbrown9 (approved):
Looks good.

PR #181, opened 2023-01-27
Clear user defaults and temp files as part of unit test setup

syoung-smallwisdom:
Once unit tests fail, could not see if I’d fixed them b/c of leftover cruft from previous test run.


Erin-Mounts (approved):
looks good

PR #183, opened 2023-01-27
Add localization keys for all caps titles

syoung-smallwisdom:
The UI/UX includes capitalizing these words when they are used in a tab view. I was handling this by having the text in the localization file be all caps on the assumption that capitalization can affect the meaning of a word. However, Android isn't doing this and the translations are confused, so until someone who speaks Spanish tells me it's a bug, then all-caps is handled in the code. ¯\\_(ツ)_/¯


Erin-Mounts (approved):
👍

PR #187, opened 2023-02-10
Add logging wrapper to iOS implementation of BridgeClient

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-255

I'd appreciate it if both of you could take a look at what I've done here. I tried to ensure that errors are only reported once in the upload code for the root cause of a failure, but there are a whole lot of them and I probably missed a few things.

Thanks!

### BridgeClientKMM 
- Added a ios-specific logger to allow sending messages to iOS so that they can be forwarded to crashlytics (or whatever).
- Changed some of the logging levels in the auth manager.

### BridgeClientExtension (Swift code)
- Added a `Logger` with some simplistic support for debug level that also includes a hook for setting the user id.
- Replaced `debugPrint()` with calls to the logger.


SwiftPackage/Sources/BridgeClientExtension/File Uploading/BridgeFileUploadManager.swift
```diff
@@ -419,14 +420,13 @@ extension BridgeFileUploadAPITyped {
     
     public func notifyBridgeUploadSucceeded(relativePath: String, uploadMetadata: BridgeFileUploadMetadataBlob) {
         guard let notifyUrl = self.notifyBridgeUrlString(for: uploadMetadata) else {
-            debugPrint("Unexpected: Could not get URL to notify Bridge of upload success from file upload metadata: \(uploadMetadata)")
```
Erin-Mounts, line 422: Did you mean to just remove this, or did you miss adding a Logger.log() call for it?

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/IOSLogger.kt
```diff
@@ -0,0 +1,78 @@
...
+        nativeLogWriter.log(
+            severity = severity.toNative(),
+            message = logMessage.ifEmpty { null },
+            tag = tag,
+            error = error
+        )
+    }
+}
+
+// It is recommended to *not* expose the Kermit logging library to iOS and instead
```
Erin-Mounts, line 57: Showing my age, I see "Kermit" and think "someone's still using that old file transfer protocol?"

Also KermitFlail.gif

Erin-Mounts (approved):
One question and one random comment, otherwise looks good to me

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BridgeFileUploadManager.swift
syoung-smallwisdom: Looks like I did mean to remove it. It's not actually an unexpected error for a participant upload. Only for a study upload. So the error logging is handled within the call to `self.notifyBridgeUrlString`.

Ideally, this would all be refactored to throw errors where a nil means "nope" to differentiate between nil as "something went wrong" and nil b/c something just isn't applicable, but that ended up being a whole big thing that I wasn't confident about tackling. ;)

nategbrown9 (approved):
Looks good.

PR #190, opened 2023-02-13
Add startedOn to upload metadata

nategbrown9:


assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
```diff
@@ -63,7 +65,8 @@ class AssessmentResultArchiveUploader(
 
         val uploadMetadata: Map<String, JsonElement> = mapOf(
             "instanceGuid" to JsonPrimitive(assessmentInstanceId),
-            "eventTimestamp" to JsonPrimitive(eventTimestamp)
+            "eventTimestamp" to JsonPrimitive(eventTimestamp),
+            "startedOn" to JsonPrimitive(startedOn.toString())
```
syoung-smallwisdom, line 69: So... does this create the timestamp in the format expected by the exporter? (Or is that old news?)

Erin-Mounts, line 69: is startedOn always available for Android uploads, or if not, how would this handle that? (iOS code I just reviewed didn't assume it would always be there)

syoung-smallwisdom (approved):
One question. Otherwise, looks good.

Note: On iOS, archives are handled generically b/c they may not be tied to an assessment (Don't remember why that was) so on iOS these properties could be nil. (I think it was for DIAN, maybe?)

Erin-Mounts (approved):
One question, otherwise looks good. Trying to think if there's a sensible way to unit-test this that isn't trivial/redundant/tautological but drawing a blank.

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentResultArchiveUploader.kt
nategbrown9: Yes, kotlinx.datetime.Instant.toString() creates an iso8601 string.

nategbrown9: Android uses AssessmentModel, which has startedOn as a non-nullable field.

PR #193, opened 2023-02-21
Fixes for mtb tests

nategbrown9:
Couple fixes to code that is not currently being used. Found the issues when writing tests for Android Mobile Toolbox app.


syoung-smallwisdom (approved):
Looks good

PR #194, opened 2023-02-21
Prep for macOS

syoung-smallwisdom:
So... it turns out that Kotlin does not currently support building a native Mac app that can be built for both Mac Silicon and Intel chips. XCFrameworks do not support both and the "work-around" is to look at the device you are building on and build for that device only.  That said, there was still some useful code clean-up that I did in order to support this. 

¯\_(ツ)_/¯ 

https://sagebionetworks.jira.com/browse/DHP-265


nategbrown9 (approved):
Looks good.

iosViewBuilder/iosViewBuilder/Assets.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,8 +1,9 @@
 {
   "images" : [
     {
-      "filename" : "Flanker 1.svg",
```
Erin-Mounts, line 4: so, does it not actually need the filename in here? was this change autogenerated?

Erin-Mounts (approved):
One question, otherwise looks good

iosViewBuilder/iosViewBuilder/Assets.xcassets/AppIcon.appiconset/Contents.json
syoung-smallwisdom: If you look, git has cleverly decided that "Arrow Matching" should be renamed "AppIcon".

I deleted "iosApp" which was autogenerated by Kotlin and wasn't set up for any useful purpose and added "iosViewBuilder" which is the naming convention we are using elsewhere for developing SwiftUI views that will not preview in a swift package. (Very annoying Xcode bug)

PR #195, opened 2023-02-22
DHP-267 Add messaging about upload and network status

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-267

Shoulda, coulda, woulda. This has been a pet-peeve for years so I decided to just do it and ask for permission later. :)

Screenshot 2023-02-22 at 10 31 10 AM
Screenshot 2023-02-22 at 10 31 32 AM
Screenshot 2023-02-22 at 10 32 09 AM
Screenshot 2023-02-22 at 10 32 26 AM
ubusercontent.com/5649217/220741479-3da6eb4a-e58b-4911-b6cf-912de8353e60.png">

### DIFF
--- a/SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
+++ b/SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
@@ -6,10 +6,22 @@
 import Foundation
 import BridgeClient
 import JsonModel
+import Network
 
 public let kPreviewStudyId = "xcode_preview"
 public let kStudyIdKey = "studyId"
 fileprivate let kUserSessionIdKey = "userSessionId"
+
+public enum NetworkStatus : String, CaseIterable {
+    /// Network connectivity is unknown
+    case unknown
+    /// The network is connected
+    case connected
+    /// The network is not connected
+    case notConnected
+    /// The user has disabled cellular
+    case cellularDenied
+}
 
 open class UploadAppManager : ObservableObject {
     
@@ -52,7 +64,19 @@ open class UploadAppManager : ObservableObject {
     @Published public var title: String
     
     /// Is the app currently uploading results or user files? This status is maintained and updated by BridgeFileUploadManager.
-    @Published public var isUploading: Bool = false
+    @Published public var isUploading: Bool = false {
+        didSet {
+            if isUploading {
+                startMonitoringNetwork()
+            }
+            else {
+                stopMonitoringNetwork()
+            }
+        }
+    }
+    
+    /// Is the app currently connected to the internet? This status is only monitored if the app is uploading data.
+    @Published public var networkStatus: NetworkStatus = .unknown
     
     /// A threadsafe observer for the `BridgeClient.UserSessionInfo` for the current user.
     public let appConfig: AppConfigObserver = .init()
@@ -280,6 +304,34 @@ open class UploadAppManager : ObservableObject {
     /// @Protected - Only this class should call this method and only subclasses should implement.
     open func updateAppState() {
         // Do nothing. Allows subclass override setup and app state changes.
+    }
+    
+    private let networkMonitoringQueue = DispatchQueue(label: "org.sagebase.NetworkConnectivityMonitor.\(UUID())")
+    private let networkMonitor: NWPathMonitor = .init()
+    
+    func stopMonitoringNetwork() {
+        networkMonitor.cancel()
+    }
+
+    func startMonitoringNetwork() {
+        networkMonitor.pathUpdateHandler = { [weak self] path in
+            guard let strongSelf = self else { return }
+            DispatchQueue.main.async {
+                switch path.status {
+                case .unsatisfied:
+                    if #available(iOS 14.2, *), path.unsatisfiedReason == .cellularDenied {
+                        strongSelf.networkStatus = .cellularDenied
+                    } else {
+                        strongSelf.networkStatus = .notConnected
+                    }
+                case .satisfied:
+                    strongSelf.networkStatus = .connected
+                default:
+                    strongSelf.networkStatus = .unknown
+                }
+            }
+         }
+        networkMonitor.start(queue: networkMonitoringQueue)
     }
     
     /// Encrypt and upload an archive created with the given builder.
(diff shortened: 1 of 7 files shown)


SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
```diff
@@ -6,11 +6,23 @@
...
 
+public enum NetworkStatus : String, CaseIterable {
+    /// Network connectivity is unknown
+    case unknown
+    /// The network is connected
+    case connected
+    /// The network is not connected
+    case notConnected
+    /// The user has disabled cellular
+    case cellularDenied
```
Erin-Mounts: Are these flags or mutually exclusive cases? I.e. can they have both disabled cellular and still be connected (via WiFi presumably)?

SwiftPackage/Sources/BridgeClientUI/Extensions/Date+Utils.swift
```diff
@@ -25,6 +25,10 @@ extension Date {
         Calendar.current.isDateInTomorrow(self)
     }
     
+    public var isThisWeek: Bool {
```
Erin-Mounts: Maybe it's just personal preference, but "this week" I usually interpret as meaning sometime between now and this coming Sunday. What about calling it 'isWithinAWeek' instead? Or is that being too pedantic...

SwiftPackage/Sources/BridgeClientUI/SingleStudy/View Models/TodayTimelineViewModel.swift
```diff
@@ -53,7 +55,10 @@ open class AbstractTodayTimelineViewModel : NSObject, ObservableObject, Schedule
                 existingSession.window = schedule
                 return existingSession
             }
-            self.sessions = filterSessions(newSessions)
+            let sessions = filterSessions(newSessions)
+            let upNext = sessions.first(where: { $0.state == .upNext })
+            self.isNextSessionSoon = upNext?.window.startDateTime.isThisWeek ?? false
```
Erin-Mounts: see above comment about isWithinAWeek (?)

SwiftPackage/Sources/BridgeClientUI/SingleStudy/Views/TodayFinishedView.swift
```diff
@@ -0,0 +1,121 @@
...
+    @State var isUploading: Bool = true
+    @State var networkStatus: NetworkStatus = .connected
+    @State var isNextSessionSoon: Bool = true
+    
+    var body: some View {
+        VStack {
+            UploadingMessageView(isUploading: $isUploading, networkStatus: $networkStatus, isNextSessionSoon: isNextSessionSoon)
+            Spacer()
+            Form {
+                Toggle("isUploading", isOn: $isUploading)
```
Erin-Mounts, line 111: IMO this UI is very much not ready for consumption by actual participants (or non-programmers in general) and shouldn't go to TestFlight looking like this lest we freak people out. It's a nice placeholder for what's possible and what we have the ability to show/message, so great for informing internal UI/UX discussions.

Erin-Mounts (commented):
Barring one comment/suggestion, the code looks good.

I didn't see anything obvious that would restrict the "buttons" parts of the UI to be seen by developers/QA only, and they're not ready for wider consumption. Also based on previous discussions with Ann and Sonia I don't think we want to stop them from leaving the app before files finish uploading unless they're in the process of logging out (and especially not if they're currently not connected to the Internet and may not be able to where they currently are). Or am I misunderstanding the application/intent of this UI? That's always a distinct possibility. Anyway I think at a minimum this will be useful leverage for getting those discussions moving and getting Design involved sooner.

SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
syoung-smallwisdom: Right, could have used an option set. Good idea.

SwiftPackage/Sources/BridgeClientUI/SingleStudy/Views/TodayFinishedView.swift
syoung-smallwisdom: This is for previewing the status header so that I can easily switch the state and check that the correct information is shown. It's not user-facing.

Erin-Mounts (approved):
Hopefully this will spur the discussion about how this should look and work so it doesn't keep getting put off. It has the potential to improve upload success/timeliness if done right. and also provides another way for us to test/see when something is wonky with the uploading process.

syoung-smallwisdom:
> Barring one comment/suggestion, the code looks good.
> 
> I didn't see anything obvious that would restrict the "buttons" parts of the UI to be seen by developers/QA only, and they're not ready for wider consumption. Also based on previous discussions with Ann and Sonia I don't think we want to stop them from leaving the app before files finish uploading unless they're in the process of logging out (and especially not if they're currently not connected to the Internet and may not be able to where they currently are). Or am I misunderstanding the application/intent of this UI? That's always a distinct possibility. Anyway I think at a minimum this will be useful leverage for getting those discussions moving and getting Design involved sooner.

So the screenshots are me being lazy and taking pictures of the *preview* that I used to be able to toggle state quickly. This header gets wired up to the `SingleStudyAppManager` and `TodayTimelineViewModel` to determine what state to show.

https://sagebionetworks.jira.com/browse/DHP-267 contains five different test case scenarios.

They can leave the app no matter what. If they aren't going to have another session "for a while" the app will tell them to "please wait". If they do not have an internet connection, the app will let them know that there is data that has not been uploaded.

Researchers can pick one **_and only one_**:
1. Get the data consistently
2. Don't tell participants what you are doing (uploading data)

Option 2 worked for mPower where the participant was asked to do something several days in a row and then was encouraged to return to the app on a semi-regular basis. For some of the Mobile Toolbox studies, this UI/UX fails spectacularly.

There are too many reasons for files to get orphaned until such time (if any) that the participant opens the app while they have an internet connection so that it will trigger an upload. Even running a scheduled background task (which we currently don't do) is only likely to fire for an app that is used *frequently*. If the app is not used frequently (for example - because their next session is 6 months later) then it is all too likely that the data never gets uploaded and neither the researcher nor the participant have any insight as to why or even know that there is anything wrong with (hypothetically) deleting the app, reinstalling it, and handing it to a *different* participant.

PR #196, opened 2023-02-23
Fix login messaging for failed network connections

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-267

While trying to create test cases for delete/reinstall scenarios, I discovered that on iOS 15.7 your phone will remember that you turned off cellular data when you went to Canada even after a delete/reinstall. Good times.

Additionally, I moved login using the `participantId:studyId` format to BridgeClientUI so that the error handling could be supported by multiple apps.


Erin-Mounts (approved):
looks good

PR #197, opened 2023-03-03
Handle http 410

nategbrown9:
Add a StateFlow that apps can observe to be notified of 410 app unsupported errors from Bridge. This will allow us to implement DHP-271 and DHP-272.

### DIFF
--- a/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
+++ b/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
@@ -11,10 +11,7 @@ import kotlinx.serialization.json.Json
 import org.koin.core.qualifier.named
 import org.koin.dsl.module
 import org.sagebionetworks.bridge.kmm.shared.BridgeConfig
-import org.sagebionetworks.bridge.kmm.shared.apis.EtagFeature
-import org.sagebionetworks.bridge.kmm.shared.apis.HttpUtil
-import org.sagebionetworks.bridge.kmm.shared.apis.RefreshTokenFeature
-import org.sagebionetworks.bridge.kmm.shared.apis.SessionTokenFeature
+import org.sagebionetworks.bridge.kmm.shared.apis.*
 import org.sagebionetworks.bridge.kmm.shared.cache.ResourceDatabaseHelper
 import org.sagebionetworks.bridge.kmm.shared.repo.AuthenticationProvider
 import org.sagebionetworks.bridge.kmm.shared.repo.AuthenticationRepository
@@ -88,6 +85,10 @@ internal fun HttpClient.appendDefaultConfig(
 
     install(EtagFeature) {
         storageCache = etagStorageCache
+    }
+
+    install(BridgeErrorStatusNotifierFeature) {
+        statusNotifier = authenticationRepository
     }
 
     install(SessionTokenFeature) {
--- a/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
+++ b/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
@@ -6,6 +6,8 @@ import io.ktor.client.plugins.*
 import io.ktor.http.*
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 import kotlinx.datetime.Clock
@@ -15,11 +17,12 @@ import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
 import org.sagebionetworks.bridge.kmm.shared.BridgeConfig
 import org.sagebionetworks.bridge.kmm.shared.apis.AuthenticationApi
+import org.sagebionetworks.bridge.kmm.shared.apis.BridgeErrorStatusNotifier
 import org.sagebionetworks.bridge.kmm.shared.cache.*
 import org.sagebionetworks.bridge.kmm.shared.cache.ResourceDatabaseHelper.Companion.APP_WIDE_STUDY_ID
 import org.sagebionetworks.bridge.kmm.shared.models.*
 
-interface  AuthenticationProvider {
+interface  AuthenticationProvider : BridgeErrorStatusNotifier {
     fun session() : UserSessionInfo?
     suspend fun reAuth() : Boolean
 }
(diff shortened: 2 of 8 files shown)


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
```diff
@@ -90,6 +87,10 @@ internal fun HttpClient.appendDefaultConfig(
         storageCache = etagStorageCache
     }
 
+    install(BridgeErrorStatusNotifierFeature) {
+        statusNotifier = authenticationRepository
+    }
```
syoung-smallwisdom, line 92: What happens when the auth calls return a 410? Then it can't get to the other calls that expect a signed in participant.

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/TestUtils.kt
```diff
@@ -36,6 +36,10 @@ data class MockAuthenticationProvider(
         userSessionInfo = userSessionInfo?.copy(sessionToken = reauthSessionToken, reauthToken = reauthToken)
         return userSessionInfo != null
     }
+
+    override fun notifyUIOfBridgeError(statusCode: HttpStatusCode) {
```
syoung-smallwisdom, line 49: This should save that it was called so that the status code can be inspected in the unit tests.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
```diff
@@ -281,11 +292,13 @@ class AuthenticationRepository(
         } ?: Pair(false, Error("reAuth token is null"))
     }
 
-    fun notifyUIOfBridgeError(statusCode: HttpStatusCode) {
-        // TODO: emm 2021-08-17 pass 410 (app version not supported) and 412 (not consented) Bridge errors along to the UI to deal with.
+    override fun notifyUIOfBridgeError(statusCode: HttpStatusCode) {
+        if (statusCode == HttpStatusCode.Gone) {
+            _appStatusMutable.value = AppStatus.UNSUPPORTED
+        }
+        // TODO: emm 2021-08-17 pass 412 (not consented) Bridge errors along to the UI to deal with.
```
syoung-smallwisdom, line 308: Remove TODO comment.

syoung-smallwisdom (changes requested):
I'd like to see a few unit tests on this. 

1. Failed auth call
2. Failed app config call (doesn't require session_id)
3. Failed call that should be authenticated

Also, the case where the user is not authenticated should still tell the app that something is very wrong and the app needs to be updated. Kind of an edge case but is it really a "don't care"?

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
nategbrown9: Good catch, this should be added to both Http clients.

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/TestUtils.kt
nategbrown9: I marked the PR as draft since I didn't yet have unit tests. Will fill this in as part of adding tests.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
nategbrown9: We aren't handling 412s yet, so I think the TODO still applies.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/di/HttpClientModule.kt
nategbrown9: Forgot that adding the AuthenticationRepository to the httpClient used for authentication creates a dependency loop (yeah we have tests that caught this). Other unauthenticated calls such as app config and study info use the other httpClient, so they will get the 410 handling.

nategbrown9: Added 410 handling to AuthenticationRepository for sign-in.

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/TestUtils.kt
nategbrown9: Now have tests

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
```diff
@@ -119,16 +119,16 @@ class AuthenticationRepository(
      * @param regionCode CLDR two-letter region code describing the region in which the phone number was issued.
      * @return Boolean
      */
-    suspend fun requestPhoneSignIn(number: String, regionCode: String) : Boolean {
+    suspend fun requestPhoneSignIn(number: String, regionCode: String) : ResourceResult<Boolean> {
```
syoung-smallwisdom, line 122: Is anyone using phone sign in?

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
```diff
@@ -221,8 +221,17 @@ class AuthenticationRepository(
         } catch (err: Throwable) {
             database.removeResource(USER_SESSION_ID, ResourceType.USER_SESSION_INFO, APP_WIDE_STUDY_ID)
             Logger.w("Error signIn: $err")
+            return ResourceResult.Failed(ResourceStatus.FAILED, getHttpStatusCode(err))
+        }
+    }
+
+    private fun getHttpStatusCode(error: Throwable) : HttpStatusCode? {
```
syoung-smallwisdom, line 228: Several of the paths are not calling this. (both reauth calls, for example). Should they be? Or are you relying upon the fact that the reauth only gets called after a call has already failed?

syoung-smallwisdom (approved):
Few questions. Otherwise, looks good.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
nategbrown9: Nope

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/AuthenticationRepository.kt
nategbrown9: I am assuming that reauth should only be called after a call has already failed, so the first call should trigger the 410.

PR #198, opened 2023-03-03
Add unit tests for reauth

syoung-smallwisdom:
- Test that reauth is called when the session token is present (and therefore getting 401 b/c it’s expired)
- Test that reauth is *not* called when the session token null (and therefore getting 401 b/c the call requires authentication)

I was looking to see if I could repro the auth calls that were being made b/c the initial call does not have a session token. Turns out that as a part of rewriting the HttpRequest config and fixing *those* errors, we also fixed this issue so my `MockBrokenAuthenticationProvider` test passes. I figure it's worth adding the tests anyway.

@dephillipsmichael 

Note: This does not address the other issues we are seeing such as the direct call to `reauth()` or why the session is returning null for a signed in user.

### DIFF
--- a/bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/HttpRequestTests.kt
+++ b/bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/HttpRequestTests.kt
@@ -6,11 +6,10 @@ import io.ktor.client.engine.mock.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.http.*
-import org.sagebionetworks.bridge.kmm.shared.BaseTest
-import org.sagebionetworks.bridge.kmm.shared.getJsonReponseHandler
-import org.sagebionetworks.bridge.kmm.shared.getTestClient
-import kotlin.test.Test
-import kotlin.test.assertEquals
+import org.sagebionetworks.bridge.kmm.shared.*
+import org.sagebionetworks.bridge.kmm.shared.models.UserSessionInfo
+import org.sagebionetworks.bridge.kmm.shared.repo.AuthenticationProvider
+import kotlin.test.*
 
 class HttpRequestTests: BaseTest() {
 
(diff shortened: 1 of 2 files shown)


bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/HttpRequestTests.kt
```diff
@@ -37,6 +36,78 @@ class HttpRequestTests: BaseTest() {
...
 
+    @Test
+    fun testRefreshToken_NullSessionToken() {
+        runTest {
+
+            var request1Headers: Headers? = null
+            var request2Headers: Headers? = null
+
+            val mockEngine = MockEngine.config {
+                // 1 - Fail first call to simulate expired token
```
nategbrown9:
```suggestion
                // 1 - Fail first call due to missing session token
```

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/HttpRequestTests.kt
```diff
@@ -37,6 +36,78 @@ class HttpRequestTests: BaseTest() {
...
+                addHandler {
+                    request1Headers = it.headers
+                    respondError(HttpStatusCode.Unauthorized)
+                }
+                // 2 successful call
+                addHandler {
```
nategbrown9, line 53: Do we need this second handler? This test is only making one http call and isn't making a call to reauth or am I missing something?

syoung-smallwisdom: I added two tests:

testRefreshToken_OldSessionToken - there is a session token, but it returns a 401 which means it's expired.

This test, where the call to `session()` is returning `null` when it shouldn't (unknown reason why) and therefore the call to reauthenticate should *not* be called, but instead the attempt to the original call should be made again without reauthenticating. This one uses the "MockBrokenAuthenticationProvider".

Both requests are made (checked below that the headers are non-null) but in this test `reauthCalled` should be false.

syoung-smallwisdom: These tests both pass b/c of changes we made to fix different issues with reauth that went into the build *after* 0.10.6, but I can't test that version of the framework (to confirm that the tests find the failure) because that version doesn't have the changes to the HttpRequest config that allow me to test that.

I figured that even those these aren't testing paths that fail the *current* code, adding tests (for regressions) wasn't a bad idea.

nategbrown9: testRefreshToken_NullSessionToken would be more clear on the expected behavior if the second handler also returned a HttpStatusCode.Unauthorized, or just have 1 handler and reuse it.

syoung-smallwisdom:
```
            val mockEngine = MockEngine.config {
                // 1 - Fail first call due to missing session token
                addHandler {
                    request1Headers = it.headers
                    respondError(HttpStatusCode.Unauthorized)
                }
                // 2 successful call
                addHandler {
                    request2Headers = it.headers
                    respond("{}", HttpStatusCode.OK, headersOf("Content-Type" to listOf(ContentType.Application.Json.toString())))
                }
                reuseHandlers = false
            }
```

Handler 1 - Fail with 401 - This is the request that gets set up with a NULL session token. 
Handler 2 - Success - This request has a valid session token that is *not* expired.

Expected behavior
1. Call is made to authRepo.session() with a NULL UserSessionInfo (faked using the MockBrokenAuthProvider)
2. Request Handler 1 responds with 401 b/c the session token is null
3. Call is made to RefreshTokenFeature which checks that (a) its a 401 and (b) calls isTokenSameOrNull
4. isTokenSameOrNull returns `false` because *now* there is a session token, but before there wasn't
5. Enter the code path where the request is tried again which updates the session token to the now-not-null value
6. Call the request (Handler 2) and get back success
7. return from the loop with the response

The expected behavior is checked with these tests:
- Hander 1 does *not* have the session token (check that the test was correctly set up)
- Handler 2 does have a session token (check that the test was correctly set up)
- That the `session()` was called 3 times
- That `reauth()` was *not* called

nategbrown9 (approved):

bridge-client/src/commonTest/kotlin/org/sagebionetworks/bridge/kmm/shared/apis/HttpRequestTests.kt
```diff
@@ -37,6 +36,103 @@ class HttpRequestTests: BaseTest() {
...
+
+            // Check assumption that the broken auth provider returns null for the first request
+            // (which throws a 401 b/c the session token is null)
+            assertNotNull(request1Headers)
+            assertNull(request1Headers!!.get("Bridge-Session"))
+
+            // Check the assumption that the broken auth provider returns non-null for the second request
+            // (which then passes b/c the session token is still valid and not expired)
+            assertNotNull(request2Headers)
+            assertEquals("testSessionToken", request1Headers!!.get("Bridge-Session"))
```
nategbrown9, line 76:
```suggestion
            assertEquals("testSessionToken", request2Headers!!.get("Bridge-Session"))
```

PR #202, opened 2023-03-18
Fix for BMC-446: Add an order column for scheduled assessments

nategbrown9:
- [x] Still need to update iOS binaries

This is a schema change to the schedule cache so need to verify migration:
- [x] iOS - Test reproing bug on broken build then updating to fixed build
- [x] android - Test reproing bug on broken build then updating to fixed build


Erin-Mounts (approved):
looks good to me

syoung-smallwisdom:
So this doesn't migrate and will crash an upgrade from build 94 on iOS.

syoung-smallwisdom:
Confirmed that migrating the views works and migration now succeeds. Thanks!

syoung-smallwisdom (approved):
Looks good. 

Need to figure out what's up with the tests. Tagged as 0.15.2 so it will link on iOS but not Android.

PR #206, opened 2023-03-29
Add logging and always update the app state

syoung-smallwisdom:
https://sagebionetworks.jira.com/browse/DHP-849

Got into a weird situation where the participant account that I was logged in as had been deleted. This caused reauth to fail and the app got stuck on the launch screen. This calls updateAppState() after the reauth is no longer pending to refresh the app state.

The bug is still open b/c that state leads to a wacky loop of infinite retry where the app will keep pinging services w/o a valid session token.


Erin-Mounts (approved):
Looks good

nategbrown9 (approved):

PR #214, opened 2023-05-15
remove debug info from upload request metadata

syoung-smallwisdom:
Apparently these fields all get added to the parquet dataset. I had originally included them to allow setting these keys in crashlytics with upload request failures.


nategbrown9 (approved):
Looks good.

PR #215, opened 2023-05-24
Add validation of metadata.json file against schema as part of test

nategbrown9:


.github/workflows/check.yml
```diff
@@ -8,13 +8,29 @@ jobs:
   build:
 
     runs-on: macos-latest
-
+    env:
+      # TODO: Figure out less fragile way of getting and setting path to jep -nbrown 5/25/13
```
Erin-Mounts: Have you tried asking ChatGPT for help with this? ¯\\\_(ツ)\_/¯ just a thought

Erin-Mounts (approved):
looks good to me.

assessmentmodel-sdk/src/test/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiverTest.kt
```diff
@@ -120,13 +123,45 @@ class AssessmentArchiverTest {
                     assertNotNull(assessmentResult.jsonSchema)
                     assertEquals(assessmentResult.jsonSchema, archiveFileInfo?.jsonSchema)
                     foundMetadataFile = true
+                    val schemaErrors = validateSchema(metadataJsonString, "https://sage-bionetworks.github.io/mobile-client-json/schemas/v1/TaskMetadata.json")
+                    assertTrue(schemaErrors.isEmpty())
```
syoung-smallwisdom: This is an old schema. Should be using https://github.com/Sage-Bionetworks/mobile-client-json/blob/main/schemas/v2/ArchiveMetadata.json

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
```diff
@@ -66,6 +66,7 @@ class AssessmentArchiver(
             appVersion = appVersion,
             deviceInfo = "${Build.PRODUCT} ${Build.MODEL}; $os",
             deviceTypeIdentifier = bridgeConfig.deviceName,
+            taskIdentifier = assessmentResult.identifier,
```
syoung-smallwisdom: This is an old schema. Latest is https://github.com/Sage-Bionetworks/mobile-client-json/blob/main/schemas/v2/ArchiveMetadata.json

syoung-smallwisdom (changes requested):

nategbrown9:
Found a java schema validator library that does what we need. Being able to run the same validator as BridgeDownstream didn't seem like it was worth the hassle of setting up Python. The other problem I found when trying to use it for the Android Mobile Toolbox App integration tests is that the python solution wont work for tests that run on a device or emulator.

syoung-smallwisdom (approved):
looks good

PR #216, opened 2023-05-30
Support archiving assessment results that adhere to Result interface

nategbrown9:
This will will allow 3rd party assessments that already generate a JSON file, to simply return a FileResult and have it archive correctly.


syoung-smallwisdom (approved):
Looks good. Found a bug in the recursive code to add a FileResult - it's always setting the content type to json, which isn't correct if we ever need to add an image or mp4.

PR #217, opened 2023-06-02
arabara/scheduled assessment adherence finish date

rabaraaaron:
https://sagebionetworks.jira.com/browse/DHP-223

Added a single query for getting the most recently finished ScheduledAssessmentAdherence, as well as a unit test for that query.

Requesting a merge into main, because develop is behind main by a lot.

### DIFF
--- a/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
+++ b/bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
@@ -146,6 +146,20 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
         return participantScheduleDatabase.getStudyBurstSchedule(studyId, timezone)
     }
 
+    /**
+     * Get the date of the most recently finished ScheduledAssessmentAdherence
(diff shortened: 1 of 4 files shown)


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/ScheduleTimelineRepo.kt
```diff
@@ -146,6 +146,20 @@ class ScheduleTimelineRepo(internal val adherenceRecordRepo: AdherenceRecordRepo
         return participantScheduleDatabase.getStudyBurstSchedule(studyId, timezone)
     }
 
+    /**
+     * Get the date of the most recently finished ScheduledAssessmentAdherence
+     * @param studyId Study identifier
+     * @param assessmentIdentifier identifier for the assessment
+     * @return the most recent completion date of this assessment, null if never completed
+     */
+    fun getLatestScheduledAssessmentAdherence(
```
syoung-smallwisdom, line 155: I think it might be better to return the record rather than just the date. That way, if the callee needs info from the `clientData` blob, they can access that. If they just need the finishedOn, they get that.

syoung-smallwisdom (approved):
Looks good. I think you should consider just returning the adherence record rather than parsing out the date.

nategbrown9 (approved):

PR #218, opened 2023-06-23
Change the assessment timeline card to a button

syoung-smallwisdom:
Tap gestures aren't easily recognized by UI testing so changing it to a button.


Erin-Mounts (approved):
Looks good

PR #219, opened 2023-06-27
If an upload request is more than a day old then cancel it and retry

syoung-smallwisdom:
These requests will always fail and will attempt to reauthenticate which will also fail.

@dephillipsmichael 

This is a band-aid to attempt to clear the orphaned files and restart after an authentication failure.

I also added some additional logging.


Erin-Mounts (approved):
looks good

dephillipsmichael:
Looks good I think.  I just want to verify that this would catch any uploads that failed to send on previous versions of the sdk, correct?

syoung-smallwisdom:
> Looks good I think. I just want to verify that this would catch any uploads that failed to send on previous versions of the sdk, correct?

Yes, there's a different (more complicated) fix that would be needed to clear the upload request.

PR #220, opened 2023-06-28
Do not attempt orphan file upload until Kotlin db is hooked up

syoung-smallwisdom:
The check for orphaned files needs to wait until after launch is done and kotlin is setup before checking for orphaned files.

Depending upon the app this can mean a whole bunch of steps that really should be sorted out *before* running the check. (MobileToolboxApp, for example, requires having a studyId set up as well as having a session token)

@dephillipsmichael This *might* fix your issue though I have another PR that I need to make on the DIAN app that may also be necessary.


Erin-Mounts (approved):
Looks good

dephillipsmichael (approved):
LGTM, it says unit tests failed tho?

syoung-smallwisdom:
> LGTM, it says unit tests failed tho?

Yeah, there's an existing issue with the Kotlin unit tests that they are failing again. This doesn't touch that code but thanks for pointing it out. I'm so use to seeing the failure that I might ignore it and not notice if it were a Swift test failure. (sigh)

PR #222, opened 2023-07-11
Part 1: Refactor upload

syoung-smallwisdom:
This does not include any changes to the iOS code related to uploading. I will need to riddle out how to set up mocks for the iOS calls (and/or figure out how to rewrite it all in Kotlin using expect/actually) but I wanted to get some eyes on this since I pulled out the easily shared code from the Android-only library.


bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/models/UploadFile.kt
```diff
@@ -23,14 +24,22 @@ data class UploadFile (
         return "uploadSession--$filePath"
     }
 
+    internal fun getSecondaryId(): String {
+        return metadata?.instanceGuid ?: filename()
+    }
+
+    internal fun filename(): String {
+        return filePath.subSequence(filePath.lastIndexOf("/")+1, filePath.length).toString()
```
Erin-Mounts, line 32: Does Kotlin not have built in methods for parsing out the components of file paths/URLs like iOS does?

Erin-Mounts (approved):
One question, otherwise looks good to me

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/models/UploadFile.kt
syoung-smallwisdom: This is copy/paste - I'm guessing it doesn't though yeah, it's a bit icky.

nategbrown9: The java File API would do this, but it is only available for Android/JVM. We now might be able to use something from Okio to do this now.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/repo/UploadRepo.kt
```diff
@@ -0,0 +1,172 @@
...
+import org.sagebionetworks.bridge.kmm.shared.cache.loadResource
+import org.sagebionetworks.bridge.kmm.shared.models.UploadFile
+import org.sagebionetworks.bridge.kmm.shared.models.UploadRequest
+import org.sagebionetworks.bridge.kmm.shared.models.UploadSession
+
+internal open class UploadRepo(
+    httpClient: HttpClient,
+    databaseHelper: ResourceDatabaseHelper,
+    backgroundScope: CoroutineScope,
+) : AbstractResourceRepo(databaseHelper, backgroundScope) {
```
nategbrown9, line 28: I don't think we gain anything by extending AbstractResourceRepo here.

nategbrown9 (approved):
Looks good.

PR #223, opened 2023-07-20
Archiver only serializes results if filename is specified

nategbrown9:


assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
```diff
@@ -103,19 +104,16 @@ class AssessmentArchiver(
...
-            }
+        if (result is BranchNodeResult) {
+            addBranchResults(result, path)
+        } else if (result is CollectionResult) {
+            recursiveAddFiles(result.inputResults, path)
+        }
+        if (result is JsonFileArchivableResult) {
+            addJsonFileArchivableResult(result, path)
+        }
+        if (result is FileResult) {
```
syoung-smallwisdom, line 162: else if?

syoung-smallwisdom (approved):
Looks good.

I think this is kinda obfuscated so some doc comments explaining how to use it might be a good idea?

ie. Why *wouldn't* you want to archive the top-level result?

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
nategbrown9: It would be a weird use case, but nothing is preventing a result from implementing both interfaces.

PR #224, opened 2023-07-21
Archive changes for 0.17.7 release

nategbrown9:
This is the same as https://github.com/Sage-Bionetworks/BridgeClientKMM/pull/223, but to a release branch based on the 0.17.6 tag.

### DIFF
--- a/assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
+++ b/assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
@@ -21,7 +21,8 @@ class AssessmentArchiver(
     private val assessmentResult: Result,
     private val jsonCoder: Json,
     private val bridgeConfig: BridgeConfig,
-    private val assessmentResultFilename: String // Used when result is of type AssessmentResult
+    // To serialize assessmentResult as part of archive, include a filename for the result.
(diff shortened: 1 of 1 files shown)


assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
```diff
@@ -103,19 +109,16 @@ class AssessmentArchiver(
...
-            }
-            is JsonFileArchivableResult -> {
-                addJsonFileArchivableResult(result, path)
-            }
-            is FileResult -> {
-                addFileResult(result, path)
-            }
+        if (result is BranchNodeResult) {
+            addBranchResults(result, path)
+        } else if (result is CollectionResult) {
```
Erin-Mounts, line 114: am I reading this change right: in this case it needs to be an either/or, but between either of these and all the other cases it's not mutually exclusive (and could fulfill one or more of them simultaneously)?

Erin-Mounts (approved):
One question, otherwise looks good to me

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
nategbrown9: BranchNodeResult implements CollectionResult, so the `else if` is to handle the rare case that we get a CollectionResult that isn't a BranchNodeResult.

PR #227, opened 2023-07-31
Refactor iOS upload - part 2

syoung-smallwisdom:
This keeps the old uploader in place. I started stubbing out the replacement but the amount of changes was getting a bit hard to grok so I'm not currently using it. A lot of this is moving code into different files so it can be used with mocks more easily. I have not yet worked out how best to send upload responses to the "correct" transfer delegate but it's getting loopy enough that I wanted you to take a look at it before I went any further. :)


SwiftPackage/Sources/BridgeClientExtension/File Uploading/ArchiveUploadProcessor.swift
```diff
@@ -0,0 +1,119 @@
...
+            return true
+        } catch {
+            Logger.log(error: error, message: "Failed to encrypt \(archive.identifier)")
+            return false
+        }
+    }
+    
+    @MainActor
+    private func _upload(archive: DataArchive, encrypted: Bool) async {
+        guard let url = archive.encryptedURL else {
```
Erin-Mounts, line 66: If we're not going to use the 'encrypted' flag here, should we maybe remove it from the function declaration (and not bother having _encrypt() return a value if we're not going to use it)?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/ArchiveUploadProcessor.swift
```diff
@@ -0,0 +1,119 @@
...
+        Logger.log(severity: .info, message: "Uploading file: \(id)", metadata: dictionary)
+        StudyDataUploadAPI.shared.upload(fileId: id, fileUrl: url, contentType: "application/zip", extras: extras)
+        do {
+            try FileManager.default.removeItem(at: url)
+        } catch let err {
+            Logger.log(error: err, message: "Failed to delete encrypted archive \(id) at \(url)")
+        }
+    }
+    
+    // TODO: syoung 07/27/2023 Replace upload with V2 uploader
```
Erin-Mounts, line 98: Isn't this what the above implementation does? Or am I missing something?

SwiftPackage/Sources/BridgeClientExtension/Data Archiving/DataArchive.swift
```diff
@@ -120,9 +120,12 @@ open class DataArchive : NSObject, Identifiable {
     
     /// Close the archive (but do not encrypt or delete).
     public final func completeArchive() throws {
+        guard !isCompleted else { return }
         let metadata = ArchiveMetadata(files: manifest)
         let metadataDictionary = try metadata.jsonEncodedDictionary()
         try addMetadata(metadataDictionary)
+        // Add info.json file b/c otherwise the upload is marked as failed. syoung 08/03/2023
```
Erin-Mounts: Per today's (2023-08-14) conversation, let's hold off on this until Dwayne's had a chance to look into what the "right" solution is. Or at least, mark it somehow or keep a note somewhere so we can be sure we'll remember to pull it back out if that solution is server-side (or a different change to the mobile apps).

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -30,44 +30,30 @@ enum HTTPMethod : String {
...
         return delegateQueue
     }()
             
-    /// The singleton background URLSession.
-    static var _backgroundSession: URLSession? = nil
+    /// The primary background URLSession.
+    var primaryBackgroundSession: (any BridgeURLSession)? = nil
     
-    /// A singleton map of pending background URLSession completion handlers that have been passed in from the app delegate and not yet called.
+    /// A map of pending background URLSession completion handlers that have been passed in from the app delegate and not yet called.
```
Erin-Mounts, line 51: Is the reason for not making this (and other things) singletons anymore just so it's possible to use mock versions for testing?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -115,21 +96,39 @@ class BackgroundNetworkManager: NSObject, URLSessionBackgroundDelegate {
...
             config.sharedContainerIdentifier = appGroupIdentifier
         }
-        
-        BackgroundNetworkManager._backgroundSession = URLSession(configuration: config, delegate: self, delegateQueue: BackgroundNetworkManager.sessionDelegateQueue)
+        return URLSession(configuration: config, delegate: self, delegateQueue: sessionDelegateQueue)
+    }
+    
+    
+    /// If set, URLSession(Data/Download)Delegate method calls received by the BackgroundNetworkManager
+    /// will be passed through to this object for further handling.
```
Erin-Mounts: since it's an array should it be "these objects"?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BackgroundNetworkManager.swift
```diff
@@ -115,21 +96,39 @@ class BackgroundNetworkManager: NSObject, URLSessionBackgroundDelegate {
...
+    var backgroundTransferHandlers: [BridgeURLSessionHandler] = []
+    
+    func registerBackgroundTransferHandler(_ handler: BridgeURLSessionHandler) {
+        backgroundTransferHandlers.append(handler)
+    }
+    
+    func backgroundTransferHandler(for task: BridgeURLSessionTask) -> BridgeURLSessionHandler? {
+        guard let handler = backgroundTransferHandlers.first(where: { $0.canHandle(task: task) })
+        else {
+            let error = BridgeUnexpectedNullError(category: .notFound, message: "Could not find a handler for URLSessionTask \(task.description).")
```
Erin-Mounts, line 127: The comment above the handlers array says "If set" which kind of gives the impression those are optional, but not having any being logged as an error seems to imply they're not actually optional?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
```diff
@@ -499,18 +487,23 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
...
-    var bridgeFileUploadApis = [String : BridgeFileUploadAPI]()
+    lazy private var bridgeFileUploadApis: [String : BridgeFileUploadAPI] = [
+        participantFileUploadAPI.apiString : participantFileUploadAPI,
+        studyDataUploadAPI.apiString : studyDataUploadAPI
+    ]
     
     /// The minimum delay before retrying a failed upload (in seconds).
     var delayForRetry: TimeInterval = 5 * 60
     
     /// BridgeFileUploadManager uses the BackgroundNetworkManager singleton to manage its background URLSession tasks.
```
Erin-Mounts:
```suggestion
    /// BridgeFileUploadManager uses a BackgroundNetworkManager instance to manage its background URLSession tasks.
```
(since it's not a singleton anymore?)

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
```diff
@@ -524,15 +517,13 @@ public class BridgeFileUploadManager: NSObject, URLSessionBackgroundDelegate {
         return UserDefaults.standard
     }()
 
-    /// Private initializer so only the singleton can ever get created.
-    private override init() {
-        netManager = BackgroundNetworkManager.shared
-        
-        self.uploadQueue = self.netManager.backgroundSession().delegateQueue
-        
+    init(netManager: BackgroundNetworkManager, appManager: UploadAppManager) {
```
Erin-Mounts, line 520: How does the code prevent multiple BridgeFileUploadManagers/BackgroundNetworkManagers/UploadAppManagers from managing the same uploads/downloads/background URLSessions since it's not a singleton anymore?

SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
```diff
@@ -189,12 +189,15 @@ open class UploadAppManager : ObservableObject {
         // If we have a session token and it is different from the old one then check orphaned files.
         if let newToken = session?.sessionToken, !newToken.isEmpty, newToken != oldToken {
             Logger.log(severity: .info, message: "Session token updated: newToken='\(newToken)', oldToken='\(oldToken ?? "")'")
-            BridgeFileUploadManager.shared.onSessionTokenChanged()
+            uploadManagerV1.onSessionTokenChanged()
```
Erin-Mounts, line 192: Why "V1"?

Erin-Mounts (approved):
Bunch of questions, but generally looks good

SwiftPackage/Sources/BridgeClientExtension/File Uploading/ArchiveUploadProcessor.swift
syoung-smallwisdom: Other than moving it, I didn't change this class. I don't remember what the original reason for allowing "encrypted=false" is, but BridgeClientKMM is not supporting it.

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BackgroundNetworkManager.swift
syoung-smallwisdom: Um, no. 

I took out the singletons because it makes unit testing and preview way simpler. Additionally, the whole thing is designed around the `UploadAppManager` being a singleton created by the app delegate. There's a whole order of operations that will cause crashes and/or silent failures if stuff isn't handled in the correct order. (For example, don't create the singleton for uploading before the IOSBridgeConfig has been initialized with things like the "app group identifier".)

This particular bit is a map and not a singleton because if the background URLSession is launched from an extension, then it tacks on a UUID to the identifier. That was getting swallowed and would never finish until 24 hours later when the orphan uploading code sorts out what uploads still need to be processed.

SwiftPackage/Sources/BridgeClientExtension/File Uploading/BackgroundNetworkManager.swift
syoung-smallwisdom: True. Changed the comment. Additionally, there *should* always be a handler?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/ArchiveUploadProcessor.swift
syoung-smallwisdom: No, it's currently still using the V1 uploading. I moved code around and changed from singletons to allow faking URLSession and URLSessionTask since Apple deprecated `init()`.

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
syoung-smallwisdom: It's internal and therefore will only be created (with proper init sequence) by the `UploadAppManager` which is itself a singleton designed to be created by the AppDelegate.

SwiftPackage/Sources/BridgeClientExtension/UploadAppManager.swift
syoung-smallwisdom: To differentiate it from V2. Basically, once there is a V2 in place that uses shared code (as much as reasonable) and a background task (to allow uploading adherence records as well as archives), it will still be a while before we remove the V1 code to handle in-flight uploads.

PR #228, opened 2023-09-06
Refactor uploads - Part 3

syoung-smallwisdom:
This has been tested on a single device with WiFi toggling.

It uses BridgeClient (cross platform Kotlin binary) to track pending uploads and to launch uploading the adherence records, uploads, and upload complete messages. It uses URLBackgroundSession to handle the S3 upload. Finally, it relies upon using BGTaskScheduler to schedule a background process (to run when the phone is sleeping and connected to wifi) to handle launching adherence record and archive uploading if there isn't a Wifi connection.


Erin-Mounts (approved):
Looks good--thanks for the detailed walk-through

nategbrown9 (approved):
Looks good.

PR #229, opened 2023-09-07
Update BridgeClientKMM to use SharedMobileUI included in AssessmentModelKMM

syoung-smallwisdom:
This uses the SharedMobileUI from AssessmentModelKMM instead of the separate repo


Erin-Mounts (approved):
👍🏻

PR #230, opened 2023-09-08
Only check for V1 orphaned uploads if there may be some

syoung-smallwisdom:
This is a little change to reduce chatter in the logs (by only checking for V1 uploads if the flag is not previously set to "finished". This means that a fresh install will check once, but a reauth recovery won't get clobbered.


Erin-Mounts (approved):
looks good

PR #231, opened 2023-09-14
[iOS] Add a local schema to building of “answers.json”

syoung-smallwisdom:
It's pretty straight-forward to add the schema to the "answers.json" file on iOS. This was already supported as a part of Sage Research and we deprecated it b/c we were going away from data parsing and the researchers wanted timestamps and question text, etc. That model doesn't lend itself well to the flattened table. ¯\_(ツ)_/¯ 

- Will take *slightly* longer to add this on Android but thought I'd put this out there as a "hey, this is what I was talking about in today's meeting". 
- For the sake of simplicity, I did not include format for answers that are date/time, url, etc. bc that's a bit more effort to support on Android.

### DIFF
--- a/SwiftPackage/Sources/BridgeClientExtension/Data Archiving/AssessmentArchiveBuilder.swift
+++ b/SwiftPackage/Sources/BridgeClientExtension/Data Archiving/AssessmentArchiveBuilder.swift
@@ -106,11 +106,28 @@ open class AssessmentArchiveBuilder : ResultArchiveBuilder {
         // Iterate through all the results within this collection and add if they are `FileArchivable`.
         try addBranchResults(assessmentResult)
         
-        // For backwards compatibility and adherence, add an "answers.json" dictionary.
+        // For assessment results that include "answer" results, create an "answers.json" file.
         if !answers.isEmpty {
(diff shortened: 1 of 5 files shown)


Erin-Mounts (approved):
looks good

SwiftPackage/Sources/BridgeClientExtension/Data Archiving/AssessmentArchiveBuilder.swift
```diff
@@ -183,6 +202,37 @@ open class AssessmentArchiveBuilder : ResultArchiveBuilder {
...
+        case .string(let value):
+            return (value, baseType)
+        case .integer(let value):
+            return (value, baseType)
+        case .number(let value):
+            return (value.jsonNumber(), baseType)
+        case .array(let value):
+            return (value.map { "\($0)" }.joined(separator: ","), .string)
+        default:
+            return nil
```
nategbrown9, line 233: This seems problematic to return nil for types not supported by the flatAnswer() function. Previously if we had a value it would get serialized into the answers file.

syoung-smallwisdom: Added "object" (though we don't actually have any currently supported question types that use this answer type).

nategbrown9 (approved):
Looks good.

PR #232, opened 2023-09-16
add "answers.json" file for Android

syoung-smallwisdom:
I also fixed a bug on the iOS implementation where sections weren't including the section in the path, which would result in identifiers clobbering each other.

This is a pretty bare-bones "build a table" implementation but it does account for the integer/double problem. It does not (on either iOS or Android) deal with formats for time/date, URL, significant digits, etc. that the fully featured `AnswerType` model object is designed to handle.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,6 +4,10 @@ on:
   pull_request:
     branches: [ main ]
 
+permissions:
+  checks: write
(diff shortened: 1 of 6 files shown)


Erin-Mounts (approved):
looks good to me

nategbrown9 (approved):
Looks good.

PR #238, opened 2023-10-30
Fix DHP-1140: Add rest of adherence fields to upload metadata

nategbrown9:


Erin-Mounts (approved):
looks good to me

syoung-smallwisdom (changes requested):
This will break iOS uploads.

Obj-c does not support Kotlin `JsonElement` objects. It also doesn't support automatic assignment to null so changes the `UploadMetadata` object must include building the XCFramework binary and fixing uploading to support it. Sorry.

bridge-client/src/commonMain/kotlin/org/sagebionetworks/bridge/kmm/shared/models/UploadMetadata.kt
```diff
@@ -19,4 +19,33 @@ data class UploadMetadata(
...
+}
+
+@Serializable
+data class UploadRequestMetadata(
+    val instanceGuid: String? = null,
+    val eventTimestamp: String? = null,
+    val startedOn: String? = null,
+    val finishedOn: String? = null,
+    val declined: Boolean? = null,
+    val clientData: JsonElement? = null
```
syoung-smallwisdom, line 32: timezone?

syoung-smallwisdom (approved):
Need to update the XCFramework binary before merging. There is a build configuration that you can use: `createSwiftPackage`

I think that we are creating tech debt here. Really, we should have an upload request take an optional `AdherenceRecord` object rather than "metadata" where the "metadata" is expected to match a subset of the `AdherenceRecord` properties. Just too many calls to services and too much room for typos. That said, I recognize that doing so is a more time-consuming refactor. ¯\_(ツ)_/¯

PR #242, opened 2023-11-08
Adherence can be null for upload

nategbrown9:
The ARC app uploads a baseline assessment that doesn't have adherence associated with it. The UploadRepo already handles this correctly, just need to allow it through the AssessmentResultArchiveUploader that Android uses.


syoung-smallwisdom (approved):
looks good

PR #243, opened 2023-11-20
Don't serialize AssessmentResults that implement JsonFileArchivableResult

nategbrown9:


assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
```diff
@@ -57,7 +57,7 @@ class AssessmentArchiver(
         recursiveAdd(assessmentResult)
 
         // Add assessment result file to archive
-        if (assessmentResult is AssessmentResult && assessmentResultFilename != null) {
+        if (assessmentResult is AssessmentResult && assessmentResult !is JsonFileArchivableResult && assessmentResultFilename != null) {
```
syoung-smallwisdom: So... why not just *not* include the assessmentResultFilename?

It seems like this is obfuscating the protocol with more special casing than it should.

syoung-smallwisdom (commented):
Comment got all messed up.

Basically, I am wondering why the calling method doesn't just use `null` for the filename?

assessmentmodel-sdk/src/main/java/org/sagebionetworks/bridge/assessmentmodel/upload/AssessmentArchiver.kt
```diff
@@ -42,6 +42,9 @@ class AssessmentArchiver(
     private var answersSchema: SimpleJsonSchema
 
     init {
+        if (assessmentResult !is AssessmentResult && assessmentResult !is JsonFileArchivableResult) {
```
syoung-smallwisdom, line 45: Can you add a comment to help future devs with why this is here?

syoung-smallwisdom (approved):
Looks good

PR #245, opened 2024-02-16
Add more logging to the V1 file uploader

syoung-smallwisdom:
Crashlytics is reporting that the V1 uploader is attempting to repeatedly upload an orphaned file with no attributes.

The device in question successfully uploaded the three files, but it looks like they were left behind and not deleted. I'm adding some additional logging to see if there is some pattern I can use to successfully mark the files as uploaded and delete them.

I figured out that files were being stored in the base directory of the "Application Support" directory and the uploader was counting those as orphaned files rather than skipping them.

### DIFF
--- a/SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
+++ b/SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
@@ -1042,11 +1042,13 @@ class BridgeFileUploadManager: SandboxFileManager, BridgeURLSessionDownloadDeleg
                     else {
                         continue
                 }
+                let uploadSuffix = "Uploads"
                 if isDirectory {
-                    if !name.hasSuffix("Uploads") {
+                    if !name.hasSuffix(uploadSuffix) {
                         dirEnumerator.skipDescendants()
                     }
-                } else {
+                } else if fileUrl.pathComponents.contains(where: { $0.hasSuffix(uploadSuffix) }) {
+                    // Exclude files at the base of the application support directory.
                     allFiles.append(fileUrl)
                 }
             }
--- a/SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
+++ b/SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
@@ -1355,21 +1355,30 @@ class BridgeFileUploadManager: SandboxFileManager, BridgeURLSessionDownloadDeleg
     }
     
     func apiInfoFromXAttrs(for filePath: String, debugMessage: String? = nil) -> (uploadApi: BridgeFileUploadAPI, fileId: String, contentType: String, extras: Codable?)? {
+        var message = "\(debugMessage ?? ""): File=\(filePath)"
         do {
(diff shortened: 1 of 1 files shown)


SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
```diff
@@ -1047,7 +1047,12 @@ class BridgeFileUploadManager: SandboxFileManager, BridgeURLSessionDownloadDeleg
                         dirEnumerator.skipDescendants()
                     }
                 } else {
+                    let countBefore = allFiles.count
                     allFiles.append(fileUrl)
+                    if allFiles.count != countBefore {
```
Erin-Mounts: Won't this always evaluate true? since allFiles is an array, it will always append fileUrl and increment the count by one?

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
```diff
@@ -1355,21 +1360,30 @@ class BridgeFileUploadManager: SandboxFileManager, BridgeURLSessionDownloadDeleg
...
     
-    func _apiInfoFromXAttrs(for filePath: String) throws -> (uploadApi: BridgeFileUploadAPI, fileId: String, contentType: String, extras: Codable?) {
+    func _tempFile(for filePath: String) throws -> (URL, [FileAttributeKey : Any]?) {
         let fullPath = self.fullyQualifiedPath(of: filePath)
         guard FileManager.default.fileExists(atPath: fullPath) else {
             let message = "Unexpected: Attempting to retrieve upload API info for temp file with invariant path '\(filePath) but temp file does not exist at '\(fullPath)"
             throw BridgeUnexpectedNullError(category: .missingFile, message: message)
         }
-        let tempFile = URL(fileURLWithPath: fullPath)
+        let attributes = try? FileManager.default.attributesOfItem(atPath: fullPath)
```
Erin-Mounts, line 1379: I'm not 100% certain, but I don't think this gets extended attributes, if that's what you were hoping to log. There's a completely separate api for xattrs. IIRC there's a class extension somewhere that implements this since I think xattrs api is Obj-C only.

Erin-Mounts (commented):
Couple questions/concerns

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
syoung-smallwisdom: yeah, I'm looking for the attributes like "when was the file last modified/created" rather than the extended attributes.

Erin-Mounts: ah ok makes sense

SwiftPackage/Sources/BridgeClientExtension/File Uploading/V1/BridgeFileUploadManager.swift
syoung-smallwisdom: whoops, good point. I misread the logic and thought it added all the files in the subdirectory.

Erin-Mounts (approved):

PR #246, opened 2024-02-22
Do not pass schedule munging failure to iOS app - just log it

syoung-smallwisdom:
This failure doesn’t really tell the partipants anything useful so showing them an alert is less useful then logging it to crashlytics.


bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/managers/NativeTimelineManager.kt
```diff
@@ -48,15 +60,14 @@ class NativeTimelineStudyBurstManager(
 
     fun observeStudyBurstSchedule() {
         scheduleJob = scope.launch {
-            if (!adherenceRecordRepo.loadRemoteAdherenceRecords(studyId)) {
-                updateFailed?.invoke()
-            }
+            !adherenceRecordRepo.loadRemoteAdherenceRecords(studyId)
```
syoung-smallwisdom, line 63: If this throws an error, then it is already being logged.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/managers/NativeTimelineManager.kt
```diff
@@ -48,15 +60,14 @@ class NativeTimelineStudyBurstManager(
...
             repo.getStudyBurstSchedule(studyId).collect { timelineResource ->
                 (timelineResource as? ResourceResult.Success)?.data?.let { schedule ->
                     viewUpdated(schedule.toNative())
                 } ?: run {
                     if (timelineResource is ResourceResult.Failed) {
-                        updateFailed?.invoke()
+                        Logger.e("Update of the `StudyBurstSchedule` failed.",
+                            Throwable("Reason Unknown."))
```
syoung-smallwisdom, line 70: The previous behavior was to pop an alert to the participant that said: "Error: An error has occurred" which was not particularly helpful. I'm not seeing anywhere that the failure is being logged so just using a general purpose try/catch here.

bridge-client/src/iosMain/kotlin/org/sagebionetworks/bridge/kmm/shared/managers/NativeTimelineManager.kt
```diff
@@ -68,20 +79,20 @@ class NativeTimelineManager(
     private val studyId: String,
     private val includeAllNotifications: Boolean,
     private val alwaysIncludeNextDay: Boolean,
-    scheduleMutator: ParticipantScheduleMutator?,
     private val viewUpdate: (NativeScheduledSessionTimelineSlice) -> Unit
-) : AbstractNativeTimelineManager(studyId, scheduleMutator) {
+) : AbstractNativeTimelineManager(studyId, null) {
```
syoung-smallwisdom, line 83: This is the TimelineManager used by Open Bridge App, which does not use the schedule mutator. Since I'm proposing that we move the logic for mutating the schedule per participant into BridgeClientKMM, I don't think we want to support the external mutator.

syoung-smallwisdom:
Nvmd. Closing - it's going to take more effort to unravel why this is being passed to the app.

